PR #60, opened 2015-01-23
Fix linting, modify base OAI harvester

erinspace:
Linting now should reject missing URLs! Modified base classes to accept more customized base urls, and added optional time between requests for providers of different sensitivities. Also allowed for properties to be pulled from the OAI header as well as from within the dc elements.

### DIFF
--- a/scrapi/base/__init__.py
+++ b/scrapi/base/__init__.py
@@ -1,7 +1,6 @@
 # Classes for scrAPI Harvesters
 from __future__ import unicode_literals
 
-import os
 import time
 from dateutil.parser import parse
 from datetime import date, timedelta
@@ -31,7 +30,6 @@ class BaseHarvester(object):
             return element
         else:
             return unicode(element, encoding=encoding)
-
 
 
 class OAIHarvester(BaseHarvester):
--- a/scrapi/base/__init__.py
+++ b/scrapi/base/__init__.py
@@ -33,6 +33,11 @@ class BaseHarvester(object):
 
 
 class OAIHarvester(BaseHarvester):
+    """ Create a harvester with a oai_dc namespace, in a date range.
+
+    Contains functions for harvesting from an OAI provider, normalizing,
+    and outputting in a way that scrapi can understand, in the most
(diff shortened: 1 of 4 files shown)


scrapi/base/__init__.py
```diff
@@ -89,16 +82,13 @@ def get_records(self, url, start_date, resump_token=''):
         print url
         data = requests.get(url)
 
-        try:
-            doc = etree.XML(data.content)
-        except etree.XMLSyntaxError:
-            return 'try a new url'
+        doc = etree.XML(data.content)
```
fabianvf: Does this error not occur, or is it just handled elsewhere?

scrapi/base/__init__.py
```diff
@@ -169,9 +159,8 @@ def get_properties(self, result, property_list):
 
         properties = {}
         for item in property_list:
```
fabianvf: Would you mind commenting this block, just so it's more apparent why you are doing what it is you are doing?

scrapi/base/__init__.py
erinspace: Yep, error no longer occurs - handled it in a better way by just accepting longer URLs if necessary!

fabianvf: ok, neat

scrapi/base/__init__.py
```diff
@@ -169,9 +159,8 @@ def get_properties(self, result, property_list):
...
-            prop = (
-                result.xpath('//dc:{}/node()'.format(item), namespaces=self.NAMESPACES) or [''])
-
+            prop = (result.xpath('//dc:{}/node()'.format(item), namespaces=self.NAMESPACES) or [''])
+            prop += (result.xpath('//ns0:{}/node()'.format(item), namespaces=self.NAMESPACES) or [''])
             if len(prop) > 1:
```
fabianvf: properties[item] = prop if len(prop) > 1 else prop[0]

erinspace:
ready for second pass!

PR #65, opened 2015-02-02
Logging, style, docstrings, code review!

erinspace:


fabianvf:
Looks good to me. @chrisseto second pass?

scrapi/base/__init__.py
```diff
@@ -110,26 +136,29 @@ def get_contributors(self, result):
         or creators in the OAI namespaces """
 
         contributors = result.xpath(
-            '//dc:contributor/node()', namespaces=self.NAMESPACES) or ['']
```
chrisseto: `['']` is a list with a singular empty string in it where `[]` is just an empty list.
Was this change intended?

scrapi/base/__init__.py
```diff
@@ -110,26 +136,29 @@ def get_contributors(self, result):
...
+            '//dc:contributor/node()',
+            namespaces=self.NAMESPACES
+        ) or []
         creators = result.xpath(
-            '//dc:creator/node()', namespaces=self.NAMESPACES) or ['']
+            '//dc:creator/node()',
+            namespaces=self.NAMESPACES
+        ) or []
```
chrisseto: See above comment.

scrapi/base/__init__.py
```diff
@@ -110,26 +136,29 @@ def get_contributors(self, result):
...
 
         all_contributors = contributors + creators
 
         contributor_list = []
         for person in all_contributors:
-            if person:
```
chrisseto: Why was this null check removed?

scrapi/base/__init__.py
```diff
@@ -171,36 +204,61 @@ def get_properties(self, result, property_list):
 
         properties = {}
         for item in property_list:
-            prop = (result.xpath('//dc:{}/node()'.format(item), namespaces=self.NAMESPACES) or [''])
-            prop += (result.xpath('//ns0:{}/node()'.format(item), namespaces=self.NAMESPACES) or [''])
-
-            properties[item] = prop if len(prop) > 1 else prop[0]
+            prop = (
```
chrisseto: Can you remove the surrounding `()` it makes the value look like it should be a tuple when its not.

scrapi/base/__init__.py
```diff
@@ -171,36 +204,61 @@ def get_properties(self, result, property_list):
...
+                result.xpath(
+                    '//dc:{}/node()'.format(item),
+                    namespaces=self.NAMESPACES
+                ) or ['']
+            )
+            prop += (
```
chrisseto: See above comment.

scrapi/base/__init__.py
```diff
@@ -171,36 +204,61 @@ def get_properties(self, result, property_list):
...
+                result.xpath(
+                    '//ns0:{}/node()'.format(item),
+                    namespaces=self.NAMESPACES
+                ) or ['']
+            )
+
+            if len(prop) > 1:
+                properties[item] = [self.copy_to_unicode(item) for item in prop]
+            else:
+                properties[item] = self.copy_to_unicode(prop[0])
```
chrisseto: This will make `properties[item]` a string rather than a list, is this intended?

scrapi/base/__init__.py
```diff
@@ -171,36 +204,61 @@ def get_properties(self, result, property_list):
...
-        return self.copy_to_unicode(title)
+        title = result.xpath(
+            '//dc:title/node()',
+            namespaces=self.NAMESPACES)
+        return self.copy_to_unicode(title[0])
 
     def get_description(self, result):
-        description = (result.xpath('//dc:description/node()', namespaces=self.NAMESPACES) or [''])[0]
-        return self.copy_to_unicode(description)
+        description = (
```
chrisseto: Same comment about tuples here

scrapi/base/__init__.py
erinspace: Right! Good point, removed, because xpath returns just an empty list with no elements found anyhow. Changed to simplify for loop below.

scrapi/base/__init__.py
erinspace: Since the list will just be empty if there are no contributors, no more need for the null check.

scrapi/base/__init__.py
erinspace: Right! Updated other things but forgot to remove the (), thank you.

scrapi/base/__init__.py
erinspace: Yep! Wanted to have singular items show up as just one instead of one item inside a list - this is done to make things as generic as possible, and because items that are singular - publisher, type, etc, will always have just one element, and it'd be strange for them to show up as just one element in a list. 

We can discuss this though!

fabianvf:
@chrisseto, I'm going to merge, let me know if I shouldn't

PR #68, opened 2015-02-10
Put parenthesis in the right place

fabianvf:
@chrisseto


scrapi/linter/util.py
```diff
@@ -35,4 +35,4 @@ def truthy(actual, expected, name):
     pretty_isinstance(actual, expected, name)
 
     if not actual:
-        raise TypeError('Expected {name} to exist, but it does not').format(name=name)
+        raise TypeError('Expected {name} to exist, but it does not'.format(name=name))
```
chrisseto: Why are you using named arguments here if there is just one?

PR #69, opened 2015-02-10
Elasticsearch processor for scrapi

fabianvf:
@chrisseto @jeffspies @erinspace 

Elasticsearch backend for scrapi. Filters data based on the FRONTEND_KEYS setting, so that the data we expose to to the users is easily configurable.

### DIFF
--- a/scrapi/processing/elastic_search.py
+++ b/scrapi/processing/elastic_search.py
@@ -1,0 +1,33 @@
+import logging
+
+from elasticsearch import Elasticsearch
+
+from scrapi import settings
+from scrapi.processing.base import BaseProcessor
+
+es = Elasticsearch(
+    settings.ELASTIC_URI,
+    request_timeout=settings.ELASTIC_TIMEOUT
+)
+
+logging.getLogger('elasticsearch').setLevel(logging.WARN)
+logging.getLogger('elasticsearch.trace').setLevel(logging.WARN)
+logging.getLogger('urllib3').setLevel(logging.WARN)
+logging.getLogger('requests').setLevel(logging.WARN)
+es.cluster.health(wait_for_status='yellow')
+
+
+class ElasticsearchProcessor(BaseProcessor):
+    NAME = 'elasticsearch'
+
+    def process_raw(*args, **kwargs):
+        pass
+
+    def process_normalized(self, raw_doc, normalized):
+        es.index(
+            body=normalized.attributes,
+            refresh=True,
+            index='share',
+            doc_type=normalized['source'],
+            id=normalized['id']['serviceID'],
+        )
--- a/scrapi/processing/elastic_search.py
+++ b/scrapi/processing/elastic_search.py
@@ -1,7 +1,7 @@
-import json
 import logging
 
 from elasticsearch import Elasticsearch
+from elasticsearch.exceptions import NotFoundError
 
 from scrapi import settings
 from scrapi.processing.base import BaseProcessor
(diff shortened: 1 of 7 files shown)


scrapi/processing/elastic_search.py
```diff
@@ -0,0 +1,37 @@
+import logging
+
+from elasticsearch import Elasticsearch
+
+from scrapi import settings
+from scrapi.processing.base import BaseProcessor
+
+es = Elasticsearch(
+    settings.ELASTIC_URI,
```
fabianvf: Why is this connection created outside of the class?

fabianvf: The class is constantly torn down and built back up, but the module is only imported once. It's relatively expensive to make a new connection, so I figured it would be better to keep it outside (just like the cassandra processor).

scrapi/processing/elastic_search.py
```diff
@@ -0,0 +1,37 @@
...
+logging.getLogger('elasticsearch.trace').setLevel(logging.WARN)
+logging.getLogger('urllib3').setLevel(logging.WARN)
+logging.getLogger('requests').setLevel(logging.WARN)
+es.cluster.health(wait_for_status='yellow')
+
+
+class ElasticsearchProcessor(BaseProcessor):
+    NAME = 'elasticsearch'
+
+    def process_raw(*args, **kwargs):
```
fabianvf: Why not just not define it?

fabianvf: Good point

scrapi/processing/elastic_search.py
```diff
@@ -0,0 +1,57 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ElasticsearchProcessor(BaseProcessor):
+    NAME = 'elasticsearch'
+
+    def process_normalized(self, raw_doc, normalized):
+        data = {
+            key: value for key, value in normalized.attributes.items()
+            if key in settings.FRONTEND_KEYS
```
chrisseto: Could this be more descriptive?

scrapi/processing/elastic_search.py
```diff
@@ -0,0 +1,57 @@
...
+            refresh=True,
+            index='share',
+            doc_type=normalized['source'],
+            id=normalized['id']['serviceID'],
+        )
+
+    def version_dateUpdated(self, normalized):
+        try:
+            old_doc = es.get_source(
+                index='share',
```
chrisseto: Can we make this a variable?

scrapi/processing/elastic_search.py
```diff
@@ -0,0 +1,57 @@
...
+                doc_type=normalized['source'],
+                id=normalized['id']['serviceID'],
+            )
+        except NotFoundError:
+            # Normally I don't like exception-driven logic,
+            # but this was the best way to handle missing
+            # types, indices and documents together
+            date = normalized['dateUpdated']
```
chrisseto: Why not just return here?

scrapi/processing/elastic_search.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def version_dateUpdated(self, normalized):
+        try:
+            old_doc = es.get_source(
+                index='share',
+                doc_type=normalized['source'],
+                id=normalized['id']['serviceID'],
+            )
+        except NotFoundError:
+            # Normally I don't like exception-driven logic,
```
chrisseto: Why the sassy comment?

scrapi/processing/elastic_search.py
fabianvf: 1. No, each document is pretty much unique (ideally)
2. At the moment you can't specify a nested field (you can include it by specifying the top level field)
3. descriptive?

scrapi/processing/elastic_search.py
```diff
@@ -0,0 +1,57 @@
...
+
+from scrapi import settings
+from scrapi.processing.base import BaseProcessor
+
+es = Elasticsearch(
+    settings.ELASTIC_URI,
+    request_timeout=settings.ELASTIC_TIMEOUT
+)
+
+logging.getLogger('elasticsearch').setLevel(logging.WARN)
```
chrisseto: Triangles

scrapi/processing/elastic_search.py
fabianvf: Had that originally, but I didn't like how the stacked returns looked. Can change it back if you want.

PR #71, opened 2015-02-11
Standardized properties as list, remove bad properties variables

erinspace:


fabianvf:
Seems good to me. Will merge in the next hour unless @chrisseto manages to get this ping and respond with a code review in time.

scrapi/base/__init__.py
```diff
@@ -211,12 +211,9 @@ def get_properties(self, result, property_list):
...
                 '//ns0:{}/node()'.format(item),
                 namespaces=self.NAMESPACES
-            ) or [''])
+            ))
 
-            if len(prop) > 1:
-                properties[item] = [self.copy_to_unicode(item) for item in prop]
-            else:
-                properties[item] = self.copy_to_unicode(prop[0])
+            properties[item] = [self.copy_to_unicode(element) for element in prop]
```
chrisseto: So `properties[item]` will always be a list?
Will this break the current elasticsearch mappings?

erinspace: I'll run it with all the consumers to make sure, but fabian said it should be fine!

chrisseto: And what was the reason for making all of these lists?

erinspace: Well, originally it was debugging for when the es mapping broke, but that wasn't the problem in the end. We decided to keep it, because we realized that trying to programmatically iterate through all of the properties, and them having inconsistent data types might be annoying (most especially when there is only one of an item in one record, and multiple in another)

chrisseto: This line has kinda bothered me for a while so:
```python
return {
    key: [
        copy_to_unicode(element)  # This should be a utility function rather than a class method
        for element in
        result.xpath('...'.format(key)) + # Consider moving these into private helpers
        result.xpath('...'.format(key))
    ]
    for key in properties_list
}
```

erinspace: Good point! Moved to util function and changed in base class!

erinspace:
Also added a timezone granularity option for good measure (arxiv_oai was breaking)

PR #85, opened 2015-02-25
Update base harvester

erinspace:
Blank titles are now set to empty string, and deleted records are not normalized (as they do not contain any information except for the service ID). Also, rejected sets are now only logged after all setSpecs in the header have been checked.

### DIFF
--- a/scrapi/base/__init__.py
+++ b/scrapi/base/__init__.py
@@ -221,7 +221,8 @@ class OAIHarvester(BaseHarvester):
     def get_title(self, result):
         title = result.xpath(
             '//dc:title/node()',
-            namespaces=self.NAMESPACES)
+            namespaces=self.NAMESPACES
+        ) or ['']
         return util.copy_to_unicode(title[0])
 
     def get_description(self, result):
@@ -262,4 +263,7 @@ class OAIHarvester(BaseHarvester):
             'dateUpdated': self.get_date_updated(result)
         }
 
+        import json
+        with open('results/{}_normalized_results.json'.format(payload['source']), 'a') as f:
+            f.write(json.dumps(payload, indent=4))
         return NormalizedDocument(payload)
--- a/scrapi/base/__init__.py
+++ b/scrapi/base/__init__.py
@@ -256,7 +256,7 @@ class OAIHarvester(BaseHarvester):
                 logger.info('Series {} not in approved list'.format(set_spec))
                 return None
 
-        payload = {
+        normalized = {
             'source': self.name,
             'title': self.get_title(result),
             'description': self.get_description(result),
@@ -267,8 +267,9 @@ class OAIHarvester(BaseHarvester):
             'dateUpdated': self.get_date_updated(result)
         }
 
-        if (result.xpath('ns0:header/@status', namespaces=self.NAMESPACES) or [''])[0] == 'deleted':
-            logger.info('Deleted record, not normalizing {}'.format(payload['id']['serviceID']))
+        status = result.xpath('ns0:header/@status', namespaces=self.NAMESPACES)
+        if status and status[0] == 'deleted':
+            logger.info('Deleted record, not normalizing {}'.format(normalized['id']['serviceID']))
             return None
 
-        return NormalizedDocument(payload)
+        return NormalizedDocument(normalized)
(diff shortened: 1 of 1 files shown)


scrapi/base/__init__.py
```diff
@@ -246,9 +247,8 @@ def normalize(self, raw_doc):
                 item_mod = item.replace('publication:', '')
                 if item_mod in self.approved_sets:
```
chrisseto: Could this be something like:
```python
if set([x.replace('publication:', '') for x in items]).intersection(self.approved_sets):
    logger.log(...)
    return None
```

scrapi/base/__init__.py
```diff
@@ -241,14 +242,8 @@ def normalize(self, raw_doc):
...
             )
-            approved = False
-            for item in set_spec:
-                item_mod = item.replace('publication:', '')
-                if item_mod in self.approved_sets:
-                    approved = True
-                else:
-                    logger.info('Series {} not in approved list'.format(item))
-            if not approved:
+            if not set([x.replace('publication:', '') for x in set_spec]).intersection(self.approved_sets):
```
fabianvf: There is a lot going on here. It might be nice to add a comment explaining what is happening in that list/set comprehension. Also, is there any reason you are doing a list comprehension and converting it to a set, rather than just a set comprehension? https://docs.python.org/2/tutorial/datastructures.html#sets

scrapi/base/__init__.py
```diff
@@ -262,4 +257,10 @@ def normalize(self, raw_doc):
             'dateUpdated': self.get_date_updated(result)
         }
 
+        if (result.xpath('ns0:header/@status', namespaces=self.NAMESPACES) or [''])[0] == 'deleted':
+            logger.info('Deleted record, not normalizing {}'.format(payload['id']['serviceID']))
+            return None
+        import json
```
fabianvf: this import should probably be at the top, right?

scrapi/base/__init__.py
```diff
@@ -262,4 +257,10 @@ def normalize(self, raw_doc):
...
+        with open('results/{}_normalized_results.json'.format(payload['source']), 'a') as f:
```
fabianvf: Why are you writing this to a file here? If you want it to be for logging/record keeping purposes, it should probably be using a function defined in the core scrapi, so that it can be consistent with all of the other information we keep track of.

erinspace: Gah! This was a total mistake, my bad. Was just in for testing! Will remove ASAP!

scrapi/base/__init__.py
chrisseto: @erinspace I'm not actually sure if `set.intersection` is the proper function.
Also apparently set comprehensions exist!
If possible could there be a test for this method?

fabianvf: It may be better to break this into a function and expand it to more lines, since it seems to be pretty difficult to reason about even at the moment

erinspace: The set intersection worked perfectly! I tested it out in ipython and with a bunch of harvesters, and this does exactly what we want. Just had to add a "not" from chris' proposed solution....

I'm happy to simplify, or break it up, or add documentation, whatever we think is best.

chrisseto: Its seems pretty straight forward IMO.
If there is not a intersection between `set_spec` and `approved_sets` don't normalize.

scrapi/base/__init__.py
```diff
@@ -241,14 +250,10 @@ def normalize(self, raw_doc):
...
-            for item in set_spec:
-                item_mod = item.replace('publication:', '')
-                if item_mod in self.approved_sets:
-                    approved = True
-                else:
-                    logger.info('Series {} not in approved list'.format(item))
-            if not approved:
+            # check if there's an intersection between the approved sets and the
+            # setSpec list provided in the record. If there isn't, don't normalize.
+            if not set([x.replace('publication:', '') for x in set_spec]).intersection(self.approved_sets):
```
chrisseto: This can just be a set comprehension as mentioned by @fabianvf 
`{x.replace('publication:', '') for x in set_spec}`

PR #98, opened 2015-02-27
Rename consumer to harvester, to match the documentation

fabianvf:


migrations.py
```diff
@@ -25,8 +25,8 @@ def migrate_from_old_scrapi():
                     with open(oldpath) as old:
                         old_json = json.load(old)
                     new_json = {
-                        'consumerVersion': old_json['consumer_version'],
```
fabianvf: This is going to be a problem

PR #102, opened 2015-03-04
XML to JSON transformations + harvester revamp

fabianvf:
Introduces a transformer class, which can perform XML -> JSON schema conversions. The transformer takes a schema, which is a dictionary of ```target``` to ```source``` mappings, where the ```source```  can either be an xpath command, an xpath command followed by a function (for post-processing), or a list of xpath commands followed by a function for post-processing. It also defines an XMLHarvester class, which defines a normalized method that simply calls ```tranform``` on a given document.


scrapi/base/__init__.py
```diff
@@ -40,6 +41,17 @@ def lint(self):
         return lint(self.harvest, self.normalize)
 
 
+class TransformerHarvester(BaseHarvester, Transformer):
```
fabianvf: Not sure that I like this, seems unnecessary to define this class in addition to the transformer

scrapi/base/__init__.py
```diff
@@ -40,6 +41,17 @@ def lint(self):
...
+
+    def __init__(self, *args, **kwargs):
+        super(TransformerHarvester, self).__init__(*args, **kwargs)
+
+    def normalize(self, raw_doc):
+        transformed = self.transform(raw_doc['doc'])
+        for key, value in transformed.items():
+            pass
```
fabianvf: This is obviously in progress, I just did a WIP commit

scrapi/base/transformer.py
```diff
@@ -0,0 +1,52 @@
+import logging
+
+from lxml import etree
+
+XML_to_JSON = 'xml_to_json'
+
+supported_transformations = [XML_to_JSON]
```
fabianvf: I don't like the way this is done either. Maybe I'm trying to make it too generic too early.

scrapi/base/transformer.py
```diff
@@ -0,0 +1,52 @@
...
+            self.transformations[target] = lambda doc: fun(doc.xpath(source, namespaces=namespaces))
+        else:
+            raise TransformationUnsupportedError
+
+    def register_transformations(self, transformations):
+        for transformation in transformations:
+            self.register_transformation(
+                transformation[0],
+                transformation[1],
+                namespaces=transformation[2],
```
fabianvf: This is clunky.

scrapi/base/transformer.py
```diff
@@ -0,0 +1,52 @@
...
+    def transform(self, doc):
+        if self._type == XML_to_JSON:
+            doc = etree.XML(doc)
+            return {
+                key: transformation(doc) for key, transformation in self.transformations.items()
+            }
+
+    def register_transformation(self, source, target, namespaces={}, fun=lambda x: x):
+        if self._type == XML_to_JSON:
+            self.transformations[target] = lambda doc: fun(doc.xpath(source, namespaces=namespaces))
```
fabianvf: This is a nice start, but it is a pretty limiting transformation

scrapi/harvesters/test_doepages/__init__.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import unicode_literals
```
fabianvf: This harvester is just a placeholder, I wanted to see what it would look like to have one. It would be nice to have another abstraction that handled harvesting, but that is a separate problem.

scrapi/harvesters/test_doepages/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+        element = ''.join(element)
+        if isinstance(element, unicode):
+            return element
+        else:
+            return unicode(element, encoding=encoding)
+
+h = DoepagesHarvester(XML_to_JSON)
+
+h.register_transformations([
+    ('//dc:title/node()', 'title', h.NAMESPACES),
```
fabianvf: This could possibly go into some kind of configuration file? The only problem is if you need to specify additional processing, that will probably be straight python code.

scrapi/base/schemas.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import unicode_literals
+
+from copy import deepcopy
+
+from nameparser import HumanName
+
+
+def default_name_parser(names):
+    names = names.split(';')
```
erinspace: Is this specific to a certain harvester?

scrapi/base/schemas.py
```diff
@@ -0,0 +1,54 @@
...
+        }
+        contributor_list.append(contributor)
+
+    return contributor_list
+
+
+def format_tags(all_tags):
+    tags = []
+    for taglist in all_tags.split(' '):
+        tags += taglist.split(',')
```
erinspace: Many tags are more than one word, will this split them all up into one word chunks?  For example, some tags recently - "child psychology",
"developmental psychology",
"social work"

scrapi/base/schemas.py
fabianvf: Yeah, I actually wanted your help with this part. I'd like to get all of the harvesters to use this system, but you have a much better understanding of what metadata each service provides. Right now the configuration in this file works for doepages only

scrapi/base/schemas.py
fabianvf: Yeah, I think it would, though it's hard to know how to split it reliably without doing this

fabianvf: I suppose if there are commas/semicolons in the tag string, we can separate them based on those instead of spaces

scrapi/base/schemas.py
```diff
@@ -0,0 +1,54 @@
...
+from copy import deepcopy
+
+from nameparser import HumanName
+
+
+def default_name_parser(names):
+    names = names.split(';')
+
+    contributor_list = []
+    for person in names:
```
chrisseto: This could be a list comprehension.

scrapi/base/schemas.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return contributor_list
+
+
+def format_tags(all_tags):
+    tags = []
+    for taglist in all_tags.split(' '):
+        tags += taglist.split(',')
+
+    return list(set([unicode(tag.lower().strip(" \n\t")) for tag in tags if tag.lower().strip(" \n\t")]))
```
chrisseto: You can just call `strip` here it trims whitespace by default. You dont need the second `lower()`

scrapi/base/schemas.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def update_schema(old, new):
+    d = deepcopy(old)
+    for key, value in new.items():
+        d[key] = value
```
chrisseto: This should handle nested dictionaries.

scrapi/base/transformer.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __init__(self, name, schema, namespaces):
+        super(XMLTransformer, self).__init__(schema)
+        self.namespaces = namespaces
+        self.NAME = name
+
+        self._processed = False
+
+    def process_schema(self):
+        if not self._processed:
```
chrisseto: is `self._processed` first defined here? Maybe make it `None` in either the class or in `__init__`

scrapi/base/transformer.py
```diff
@@ -0,0 +1,70 @@
...
+class XMLTransformer(BaseTransformer):
+
+    def __init__(self, name, schema, namespaces):
+        super(XMLTransformer, self).__init__(schema)
+        self.namespaces = namespaces
+        self.NAME = name
+
+        self._processed = False
+
+    def process_schema(self):
```
chrisseto: Is there a reason this isn't in `BaseTransformer`?

scrapi/base/transformer.py
```diff
@@ -0,0 +1,70 @@
...
+                schema[key] = self._process_schema(value)
+            elif isinstance(value, list) or isinstance(value, tuple):
+                schema[key] = partial(self._process_iter, value)
+            elif isinstance(value, basestring):
+                schema[key] = partial(self._process_string, value)
+        return schema
+
+    def _process_string(self, string, doc):
+        val = doc.xpath(string, namespaces=self.namespaces)
+        return '' if not val else val[0] if len(val) == 1 else val
```
chrisseto: I think the original harvester would use `['']` over `''`. ping @erinspace

scrapi/base/transformer.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def _process_iter(self, l, doc):
```
chrisseto: This could be moved to `BaseTransformer`

scrapi/harvesters/doepages/__init__.py
```diff
@@ -1 +1,94 @@
...
+        start_date = date.today() - timedelta(days_back)
+        base_url = 'http://www.osti.gov/pages/pagesxml?nrows={0}&EntryDateFrom={1}'
+        url = base_url.format('1', start_date.strftime('%m/%d/%Y'))
+        initial_data = requests.get(url)
+        print(initial_data.url)
+        record_encoding = initial_data.encoding
+        try:
+            initial_doc = etree.XML(initial_data.content)
+        except etree.XMLSyntaxError as e:
+            print("error in namespaces: {}".format(e))
```
chrisseto: Dont use `print` here. Shouldn't this just raise an exception?

scrapi/harvesters/doepages/__init__.py
```diff
@@ -1 +1,94 @@
...
+                'doc': record,
+                'source': self.NAME,
+                'docID': self.copy_to_unicode(doc_id),
+                'filetype': 'xml'
+            }))
+
+        return xml_list
+
+    def copy_to_unicode(self, element):
+        encoding = 'UTF-8'
```
chrisseto: This should be a constant or kwarg or just passed straight into the function.

tests/test_transformer.py
```diff
@@ -0,0 +1,47 @@
...
+
+class TestHarvester(XMLHarvester):
+
+    def harvest(self, days_back=1):
+        return [RawDocument({
+            'doc': str(TEST_XML_DOC),
+            'source': 'TEST',
+            'filetype': 'XML',
+            'docID': "1"
+        })] * days_back
```
chrisseto: prefer `[... for _ in xrange(number)]`

scrapi/base/transformer.py
```diff
@@ -0,0 +1,55 @@
...
+    def transform(self, doc):
+        return self._transform(self.schema, doc)
+
+    def _transform(self, schema, doc):
+        transformed = {}
+        for key, value in schema.items():
+            if isinstance(value, dict):
+                transformed[key] = self._transform(value, doc)
+            elif isinstance(value, list) or isinstance(value, tuple):
+                transformed[key] = self._process_iter(value, doc)
```
chrisseto: rename to `_transform_iter`

scrapi/base/transformer.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @abc.abstractmethod
+    def _process_string(self, string, doc):
+        raise NotImplementedError
+
+class XMLTransformer(BaseTransformer):
+
+    def __init__(self, name, schema, namespaces):
+        super(XMLTransformer, self).__init__(schema)
+        self.namespaces = namespaces
```
chrisseto: These should be abstract properties.

scrapi/base/transformer.py
```diff
@@ -0,0 +1,55 @@
...
+        self.NAME = name
+
+        self._processed = False
```
chrisseto: Is this used?

scrapi/base/__init__.py
```diff
@@ -30,17 +31,25 @@ class BaseHarvester(object):
 
     @abc.abstractmethod
     def harvest(self, days_back=1):
-        pass
+        raise NotImplementedError
```
JeffSpies: I don't think you need these. I suppose it's explicit, but I don't think we'll ever get there.

tests/utils.py
```diff
@@ -49,3 +51,99 @@
     '_id': 'yes! yes! yes!',
     'count': 0
 }
+
+
+TEST_SCHEMA = update_schema(BASEXMLSCHEMA, {
+    "properties": {
+        "title1": "//dc:title/node()",
+        "title2": ["//dc:title/node()", lambda x: x.lower()],
+        "title3": ["//dc:title/node()", "//dc:title/node()", lambda x, y: x + y.lower()]
```
JeffSpies: I'd rather these be tuples: ((args, kwargs), callable) where args or kwargs could be none and the first element would not necessarily be a tuple.

scrapi/base/__init__.py
fabianvf: I actually added these for coverage testing, so that it will ignore all lines that have ```raise NotImplementedError``` I suppose I could ignore pass instead, but I thought that had a higher chance of being real code.

tests/utils.py
fabianvf: what would the function of kwargs be?

JeffSpies: Named selectors. Good for code-as-documentation.

fabianvf: I think the downside would be that specifying named selectors would be too verbose to do inline, which would complicate the schema specification. I'm not sure that named selectors would necessarily make the code more readable.

This is what I imagine the usage looking like, and in both situations I think just having args is still simpler to read.

```python

def process_title(t, title=None):
    return t.lower() + title.lower() if title else t.lower()

title2_kwargs = {
    "title": "//dc:title/node()"
}

SCHEMA = update_schema(BASEXMLSCHEMA, {
    "properties": {
        "title1": (("//dc:title/node(),), {"title1": "//dc:title/node()"}), process_title),
        "title2": (("//dc:title/node(),), title2_kwargs), process_title)
    }
)
```

scrapi/base/schemas.py
```diff
@@ -0,0 +1,96 @@
...
+            'suffix': name.suffix,
+            'email': '',
+            'ORCID': ''
+        }
+        contributor_list.append(contributor)
+
+    return contributor_list
+
+
+def format_tags(all_tags):
```
erinspace: I'd modify this function to more match something like what we used to use in the old class based harvester for dealing with tags - instead of splitting on spaces (which would remove meaning from lots of the multi-word tags) you could split on commas, something like this: 

```
    def format_tags(all_tags):
        for tag in all_tags:
            if ', ' in tag:
                tags.remove(tag)
                tags += tag.split(',')

        return [util.copy_to_unicode(tag.lower().strip()) for tag in tags]
```

scrapi/base/schemas.py
```diff
@@ -0,0 +1,96 @@
...
+    return d
+
+
+BASEXMLSCHEMA = {
+    "description": ('//dc:description/node()', lambda x: unicode(x.strip())),
+    "contributors": ('//dc:creator/node()', lambda x: default_name_parser(x.split(';'))),
+    "title": ('//dc:title/node()', lambda x: unicode(x.strip())),
+    "dateUpdated": ('//dc:dateEntry/node()', lambda x: unicode(x.strip())),
+    "id": {
+        "url": ('//dcq:identifier-citation/node()', lambda x: unicode(x.strip())),
```
erinspace: We might want to change this line in particular to something OAI specific - this namespace is only used in doepages

tests/utils.py
JeffSpies: Then most people wouldn't have to use them, but we need to support them.
Talk in person for more.

fabianvf: I added them in, but I'm still not sure what I think about them. Perhaps we, @chrisseto and @erinspace could chat about it tomorrow. I think the ```((args, kwargs), callable)``` pattern doesn't work well from a usability perspective (after having written the test cases for it)

scrapi/base/transformer.py
```diff
@@ -0,0 +1,82 @@
...
+        if isinstance(l[0], tuple) and len(l) == 2:
+            return self._transform_arg_kwargs(l, doc)
+
+        for value in l:
+            if isinstance(value, basestring):
+                docs.append(self._transform_string(value, doc))
+            elif callable(value):
+                return value(*[res for res in docs])
+
+    def _transform_arg_kwargs(self, l, doc):
```
chrisseto: `args_kwargs`?

scrapi/base/__init__.py
```diff
@@ -66,13 +75,45 @@ class OAIHarvester(BaseHarvester):
...
+                    self.resolve_property
+                ) for item in self.property_list
+            }
+        }
+        return update_schema(OAISCHEMA, properties)
+
+    def resolve_property(self, dc, ns0):
+        if isinstance(dc, list) and isinstance(ns0, list):
+            ret = dc.extend(ns0)
+            return [val for val in ret if ret]
```
chrisseto: did you mean `[val for val in ret if val]`?

scrapi/base/__init__.py
```diff
@@ -66,13 +75,45 @@ class OAIHarvester(BaseHarvester):
...
+                item: (
+                    '//dc:{}/node()'.format(item),
+                    '//ns0:{}/node()'.format(item),
+                    self.resolve_property
+                ) for item in self.property_list
+            }
+        }
+        return update_schema(OAISCHEMA, properties)
+
+    def resolve_property(self, dc, ns0):
```
chrisseto: Better names please.

scrapi/base/__init__.py
```diff
@@ -66,13 +75,45 @@ class OAIHarvester(BaseHarvester):
...
+        if isinstance(dc, list) and isinstance(ns0, list):
+            ret = dc.extend(ns0)
+            return [val for val in ret if ret]
+        elif not dc:
+            return ns0
```
chrisseto: If `None` and `None` is passed `None` is returned. Is that expected?

scrapi/base/__init__.py
```diff
@@ -66,13 +75,45 @@ class OAIHarvester(BaseHarvester):
...
+        elif not ns0:
+            return dc
+        else:
+            return [dc, ns0]
+
     def harvest(self, days_back=1):
 
         start_date = str(date.today() - timedelta(int(days_back)))
```
chrisseto: Is there a reason to be casting `days_back` here?

chrisseto: why is `str` being used?

scrapi/base/helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    if isinstance(all_tags, basestring):
+        tags = all_tags.split(sep)
+    elif isinstance(all_tags, list):
+        for tag in all_tags:
+            if sep in tag:
+                tags.extend(tag.split(sep))
+            else:
+                tags.append(tag)
+
+    return list(set([unicode(tag.lower().strip()) for tag in tags if tag.lower().strip()]))
```
chrisseto: Use a set comprehension here

chrisseto: You dont need the second call to `lower`

scrapi/base/helpers.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def oai_extract_doi(identifiers):
+    identifiers = [identifiers] if isinstance(identifiers, basestring) else identifiers
```
chrisseto:
```python
identifiers = [identifies] if not isinstance(identifiers, list) else identifies
```

scrapi/base/helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    for item in identifiers:
+        if 'doi' in item.lower():
+            return unicode(item.replace('doi:', '').replace('DOI:', '').replace('http://dx.doi.org/', '').strip())
+    return ''
+
+
+def oai_extract_url(identifiers):
+    identifiers = [identifiers] if isinstance(identifiers, basestring) else identifiers
```
chrisseto: See above comment.

scrapi/base/helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    for arg in args:
+        if isinstance(arg, list):
+            for name in arg:
+                names.append(name)
+        elif arg:
+            names.append(arg)
+    return default_name_parser(names)
+
+
+def pack(*args, **kwargs):
```
chrisseto: :+1:

scrapi/base/transformer.py
```diff
@@ -0,0 +1,82 @@
+from __future__ import unicode_literals
+
+import abc
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class BaseTransformer(object):
```
chrisseto: Any reason to not merge this and BaseHarvester?

scrapi/base/transformer.py
```diff
@@ -0,0 +1,82 @@
...
+
+    __metaclass__ = abc.ABCMeta
+
+    def transform(self, doc):
+        return self._transform(self.schema, doc)
+
+    def _transform(self, schema, doc):
+        transformed = {}
```
chrisseto: Why is this not just `transform` if `transform` just calls `_transform`

scrapi/base/transformer.py
```diff
@@ -0,0 +1,82 @@
...
+        for key, value in schema.items():
+            if isinstance(value, dict):
+                transformed[key] = self._transform(value, doc)
+            elif isinstance(value, list) or isinstance(value, tuple):
+                transformed[key] = self._transform_iter(value, doc)
+            elif isinstance(value, basestring):
+                transformed[key] = self._transform_string(value, doc)
+        return transformed
+
+    def _transform_iter(self, l, doc):
```
chrisseto: Can this be `_transform_iterable`? `_transform_iter` sounds like it should return a generator.

scrapi/harvesters/clinicaltrials/harvester.py
```diff
@@ -13,293 +13,149 @@
...
+                count += 1
+                if count % 100 == 0:
+                    logger.info("You've requested {} studies, keep going!".format(official_count))
+                    count = 0
+                time.sleep(1)
+
+        return xml_list
+
+    @property
+    def schema(self):
```
chrisseto: This would be better suited as a class variable.

scrapi/harvesters/clinicaltrials/harvester.py
```diff
@@ -13,293 +13,149 @@
...
+        return {
+            "contributors": ('//overall_official/last_name/node()', default_name_parser),
+            "id": {
```
chrisseto: Use single quotes

scrapi/harvesters/doepages/__init__.py
```diff
@@ -1 +1,98 @@
...
+    def namespaces(self):
+        return {
+            'rdf': 'http://www.w3.org/1999/02/22-rdf-syntax-ns#',
+            'dc': 'http://purl.org/dc/elements/1.1/',
+            'dcq': 'http://purl.org/dc/terms/'
+        }
+
+    @property
+    def schema(self):
+        return update_schema(
```
chrisseto: Maybe call this method `updated_schema` to clarify that it does in fact return a new value.

scrapi/settings/harvesterManifests/test_doepages.json
```diff
@@ -0,0 +1,10 @@
+{
+    "days": "mon-sun",
+    "hour": 23,
+    "minute": 59,
+    "longName": "Department of Energy Pages",
+    "shortName": "test_doepages",
```
chrisseto: Remove the test if we are going to use this harvester

tests/test_transformer.py
```diff
@@ -0,0 +1,85 @@
...
+        args = ("//dc:title/node()", )
+        kwargs = {"title1": "//dc:title/node()"}
+
+        self.harvester.SCHEMA = update_schema(
+            TEST_SCHEMA,
+            {
+                'title': ((args, kwargs), process_title),
+                'properties': {
+                    'title2': ((args,), process_title),
+                    'title3': ((kwargs, ), process_title2),
```
chrisseto: This syntax looks odd...

tests/utils.py
```diff
@@ -1,5 +1,8 @@
 from __future__ import unicode_literals
 
+from scrapi.base.helpers import update_schema
+from scrapi.base.schemas import BASEXMLSCHEMA
+
 RAW_DOC = {
     'doc': str('{}'),
```
chrisseto: json.dumps?

scrapi/base/__init__.py
fabianvf: yes

scrapi/base/__init__.py
fabianvf: not sure, but this PR does not touch the harvest method. @erinspace

scrapi/base/helpers.py
fabianvf: I think the logic would be difficult to capture in a set comprehension.

scrapi/base/transformer.py
fabianvf: _transform takes a schema which it can recurse on, you can't recurse on self.schema

scrapi/base/__init__.py
fabianvf: dc corresponds to the dc namespace, ns0 to the ns0 namespace. What would you prefer they be named?

scrapi/base/transformer.py
fabianvf: I suppose not, though the function of the classes are different enough that it may be easier to understand with them separate. I don't really have a strong opinion one way or the other though.

scrapi/base/__init__.py
erinspace: originally str was used to correctly format the date for insertion into the provider API call

scrapi/harvesters/doepages/__init__.py
```diff
@@ -1 +1,98 @@
...
+            'dcq': 'http://purl.org/dc/terms/'
+        }
+
+    @property
+    def schema(self):
+        return updated_schema(
+            BASEXMLSCHEMA,
+            {
+                "properties": {
+                    "language": '//dc:language/node()',
```
chrisseto: Use single quotes

fabianvf: "don't tell me how to live my life''
-'Fabian"

PR #110, opened 2015-03-11
Registry for harvesters + remove manifest system

chrisseto:


scrapi/__init__.py
```diff
@@ -1 +1,32 @@
...
+
+    @property
+    def beat_schedule(self):
+        return {
+            'run_{}'.format(name): {
+                'args': [name],
+                'schedule': crontab(**inst.run_at),
+                'task': 'scrapi.tasks.run_harvester',
+            }
+            for name, inst
```
fabianvf: Minor thing, but can inst be called harvester or something instead? Just a little more clear

scrapi/base/__init__.py
```diff
@@ -38,6 +58,14 @@ def normalize(self, raw_doc):
     def lint(self):
         return lint(self.harvest, self.normalize)
 
+    @property
```
fabianvf: Does this need to be a property?

scrapi/base/__init__.py
```diff
@@ -65,14 +93,20 @@ class OAIHarvester(BaseHarvester):
...
-        self.approved_sets = approved_sets
+    def __init__(self, timezone_granularity=False, timeout=0.5, property_list=None, approved_sets=None):
+        # self.name = name
+        # self.base_url = base_url
         self.timeout = timeout
         self.timezone_granularity = timezone_granularity
 
+    @property
+    def property_list(self):
+        return ['date', 'language', 'type']
```
fabianvf: This is probably just a WIP thing, but just noting to make sure this doesn't stay

scrapi/base/__init__.py
```diff
@@ -65,14 +93,20 @@ class OAIHarvester(BaseHarvester):
...
+
+    @property
+    def file_format(self):
```
fabianvf: does this need to be a property?

scrapi/harvesters/__init__.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+_, __all__, files = next(os.walk(os.path.dirname(__file__)))
+
+for name in files:
+    if name[-3:] == '.py' and name != '__init__.py':
+        __all__.append(name[:-3])
+
```
fabianvf: wow

scrapi/harvesters/mit.py
```diff
@@ -0,0 +1,891 @@
+
```
fabianvf: no newline here

scrapi/harvesters/mit.py
```diff
@@ -0,0 +1,891 @@
...
+"""
+
+
+from __future__ import unicode_literals
+
+from scrapi.base import OAIHarvester
+
+
+class MITHarvester(OAIHarvester):
+    long_name = 'MIT'
```
fabianvf: Shouldn't it be Massachusetts Institute of Technology?

scrapi/base/__init__.py
chrisseto: Why? This is what the defaults where in the OAI harvester.

scrapi/base/__init__.py
chrisseto: Possibly? We should certainly space them out but that could be auto incremented by the registry.

scrapi/base/__init__.py
fabianvf: My bad, I missed that the original defined defaults.

scrapi/harvesters/__init__.py
fabianvf: And not in a bad way, though maybe add some commentary, since there's a good chance someone could be completely mystified by this while writing a harvester.

scrapi/base/__init__.py
fabianvf: Ah, I see. Smart

scrapi/harvesters/mit.py
```diff
@@ -0,0 +1,891 @@
...
+    short_name = 'mit'
+    base_url = 'http://dspace.mit.edu/oai/request'
+    property_list = [
```
erinspace: Is there a reason why approved sets is a python property, but property_list isn't?

scrapi/harvesters/mit.py
erinspace: Should really be DSpace@MIT - http://dspace.mit.edu/

scrapi/base/__init__.py
```diff
@@ -20,13 +21,32 @@
 logger = logging.getLogger(__name__)
 
 
+class HarvesterMeta(abc.ABCMeta):
+    def __init__(cls, name, bases, dct):
+        super(HarvesterMeta, cls).__init__(name, bases, dct)
+        if not isinstance(cls.short_name, abc.abstractproperty):
```
fabianvf: This should throw an error if it isn't there, otherwise we can miss harvesters (like what happened with ucescholarship)

scrapi/harvesters/__init__.py
fabianvf: docstrings

scrapi/processing/base.py
```diff
@@ -2,7 +2,7 @@ class BaseProcessor(object):
     NAME = None
 
     def process_raw(self, raw_doc, **kwargs):
-        pass
+        pass  # pragma: no cover
```
fabianvf: Why not make them raise NotImplementedErrors?

tests/test_base.py
```diff
@@ -0,0 +1,116 @@
...
+                file_format = 'test'
+                normalize = lambda x: x
+
+            TestHarvester()
+
+        assert e.value.message == self.ERR_MSG.format('harvest')
+
+    def test_requires_normalize(self):
+        with pytest.raises(TypeError) as e:
+            class TestHarvester(BaseHarvester):
```
fabianvf: Why define this same class so many times?

PR #113, opened 2015-03-12
Generate provider map

fabianvf:
Generates a provider map on the scrAPI side, so that this logic can be removed from the OSF. Includes favicons. Relies on registry PR.


scrapi/harvesters/arxiv.py
```diff
@@ -11,8 +11,9 @@
 
 
 class ArxivHarvester(OAIHarvester):
-    short_name = 'arxiv_oai'
-    long_name = 'arXiv'
+    short_name = 'arxiv'
```
chrisseto: This will require a migration

scrapi/harvesters/cmu.py
```diff
@@ -15,6 +15,7 @@
 class CMUHarvester(OAIHarvester):
     short_name = 'cmu'
     long_name = 'Carnegie Mellon University'
+    url = 'Carnegie Mellon University Research Showcase'
```
chrisseto: This isn't a url

scrapi/settings/travis-dist.py
```diff
@@ -13,6 +13,8 @@
 
 USE_FLUENTD = False
 
+PLOS_API_KEY = None
```
chrisseto: No longer relevant.

scrapi/harvesters/arxiv.py
fabianvf: right, good point

PR #116, opened 2015-03-17
Implement cassandra diffing

fabianvf:


scrapi/processing/cassandra.py
```diff
@@ -43,17 +43,22 @@ def send_to_database(self, docID, source, **kwargs):
...
-            return document.update(versions=versions, **kwargs)
+                return document.update(versions=versions, **kwargs)
+            else:
+                raise events.Skip("No changees detected for document with ID {0} and source {1}.".format(docID, source))
         else:
             # create document
             return DocumentModel.create(docID=docID, source=source, **kwargs)
 
+    def different(self, old, new):
+        return False in [new[key] == old[key] for key in new.keys() if key != 'timestamps']
```
chrisseto: You could use the `all` function here or just a for loop that would return `False` which would be a tad bit faster.

PR #119, opened 2015-03-23
Add compiled regex to url finder for oai harvesters

erinspace:
fixes CenterForOpenScience/osf.io#2278

PR #120, opened 2015-03-23
Add regex to url finder helper method, upgraded to hotfix

erinspace:
fixes CenterForOpenScience/osf.io#2278, upgrades #119 to a hotfix

### DIFF
--- a/scrapi/base/helpers.py
+++ b/scrapi/base/helpers.py
@@ -4,7 +4,7 @@ import re
 from copy import deepcopy
 from nameparser import HumanName
 
-URL_REGEX = re.compile(ur'(https?://\S*\/\S*)')
+URL_REGEX = re.compile(ur'(https?://\S*\.\S*)')
 
 
 def updated_schema(old, new):
--- a/scrapi/base/helpers.py
+++ b/scrapi/base/helpers.py
@@ -1,7 +1,10 @@
 from __future__ import unicode_literals
 
+import re
 from copy import deepcopy
 from nameparser import HumanName
+
+URL_REGEX = re.compile(ur'(https?://\S*\/\S*)')
 
 
 def updated_schema(old, new):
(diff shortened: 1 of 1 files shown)


scrapi/base/helpers.py
```diff
@@ -1,8 +1,11 @@
 from __future__ import unicode_literals
 
+import re
 from copy import deepcopy
 from nameparser import HumanName
 
+URL_REGEX = re.compile(ur'(https?://\S*\/\S*)')
```
chrisseto: Any reason why the middle forward slash is escaped but the first two are not?
This will also only match urls that have text after the first slash, is that expected?

scrapi/base/helpers.py
```diff
@@ -57,8 +60,12 @@ def oai_extract_doi(identifiers):
 def oai_extract_url(identifiers):
     identifiers = [identifiers] if not isinstance(identifiers, list) else identifiers
     for item in identifiers:
-        if 'http://' in item or 'https://' in item and 'viewcontent' not in item:
-            return unicode(item)
+        try:
+            found_url = URL_REGEX.search(item).group()
+            if 'viewcontent' not in found_url:
+                return found_url.decode('utf-8')
+        except AttributeError:
```
chrisseto: What raises an attribute error here?

erinspace: since this goes through every identifier tag looking for a url, .group() will raise an attribute error if it doesn't find a url in that particular identifier!

scrapi/base/helpers.py
fabianvf: should probably be ```(https?://\S*\.\S*)```.

chrisseto:
lgtm :honey_pot: @fabianvf?

fabianvf:
:+1: Only comment is that having the URL regex as an external constant to the function may be unnecessary

chrisseto:
It stops the regex from being compiled over and over again, a small optimization.

fabianvf:
Makes sense

PR #124, opened 2015-03-24
Processing -> Storage

fabianvf:


chrisseto:
Can https://github.com/fabianvf/scrapi/pull/123 be closed?

fabianvf:
yeah, if you want them to both be in the same PR. May want to just merge #123 instead though, then this diff will just get smaller. Not sure why I branched off #123 here.

chrisseto:
Could you redefine `BaseProcessor`
Would backend be a better name for this/these module(s)?

fabianvf:
We can continue this discussion later.

PR #127, opened 2015-03-26
Add constants and functions to the transformer language

fabianvf:
@erinspace @chrisseto

### DIFF
--- a/scrapi/base/schemas.py
+++ b/scrapi/base/schemas.py
@@ -10,7 +10,7 @@ from .helpers import (
     oai_process_contributors
 )
 
-SKIP = None
+CONSTANT = lambda x: lambda doc: x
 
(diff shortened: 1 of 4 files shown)


scrapi/base/schemas.py
```diff
@@ -10,6 +10,8 @@
     oai_process_contributors
 )
 
+CONSTANT = lambda x: lambda *_, **__: x
```
chrisseto: Shouldn't this only get passed doc?

chrisseto:
Could you also add a test for this functionality?

scrapi/base/schemas.py
fabianvf: I just figured this would prevent us from having to keep CONSTANT in sync with however we end up processing functions.

PR #129, opened 2015-03-26
Prototype JSON transformer with working harvester

fabianvf:


scrapi/base/transformer.py
```diff
@@ -39,16 +39,20 @@ def _transform(self, schema, doc):
         return transformed
 
     def _transform_iterable(self, l, doc):
-        docs = []
 
         if isinstance(l[0], tuple) and len(l) == 2:
             return self._transform_args_kwargs(l, doc)
 
-        for value in l:
+        fn = l[-1]
```
chrisseto:
```python
_transform_iterable(self, values, doc):
...
fn = values.pop(-1)
...
```

fabianvf: atm values can either be list or tuple

chrisseto:
```python
fn, values = l[-1], l[:-1]
```

Either way is fine, I forgot about the tuple part.

scrapi/base/transformer.py
```diff
@@ -39,16 +39,19 @@ def _transform(self, schema, doc):
...
+        fn, values = l[-1], l[:-1]
+        args = []
+
+        for value in values:
             if isinstance(value, basestring):
-                docs.append(self._transform_string(value, doc))
+                args.append(self._transform_string(value, doc))
             elif callable(value):
-                return value(*[res for res in docs])
+                args.append(value(doc))
```
chrisseto: Why was this changed to an append?

previously:
```python
if isinstance(...):
    args.append(...)
elif callable(...):
    return ...
```
currently:
```python
if isinstance(...):
    args.append(...)
elif callable(...):
    args.append(...)
return ...
```

Won't these behaviors be very different?

scrapi/harvesters/biomed.py
```diff
@@ -15,20 +15,73 @@
...
+from scrapi.base.schemas import CONSTANT
+from scrapi.linter.document import RawDocument
 
 logger = logging.getLogger(__name__)
 
 
-class BiomedHarvester(BaseHarvester):
+def process_contributors(authors):
+
+    authors = authors.strip().replace('<span class="author-names">', '').replace('</span>', '')
```
chrisseto: `strip` may be a better option here if these are only at the beginning and then end.

My uncle `</span>` has been trying to get published in biomed for a while now...

scrapi/harvesters/biomed.py
```diff
@@ -15,20 +15,73 @@
...
+    authors = authors.split(',')
+
+    if ' and ' in authors[-1] or ' <em>et al.</em>' in authors[-1]:
+        split_name = authors.pop(-1).replace(' <em>et al.</em>', '').split(' and ')
+        authors.extend(split_name)
+
+    contributor_list = []
+    for person in authors:
```
chrisseto: This couldn't be a list comprehension.

scrapi/harvesters/biomed.py
```diff
@@ -15,20 +15,73 @@
...
     url = 'http://www.biomedcentral.com/'
+    count = 0
 
-    file_format = 'json'
     URL = 'http://www.biomedcentral.com/search/results?terms=*&format=json&drpAddedInLast={}&itemsPerPage=250'
 
+    schema = {
+        'title': ('bibliographyTitle', 'blurbTitle', lambda x, y: x or y),
+        'description': 'blurbText',
+        'tags': CONSTANT([]),
```
chrisseto: :+1:

scrapi/base/transformer.py
fabianvf: yes, the new behavior is better. With the old behavior, it would return if it found a function, so you could only ever include a single function. Now you can include a list of functions that can be called on docs, and the last function will be called with the results of the previous functions. It is now more consistent with how the strings are processed, and I think it adds a lot of flexibility. It should cause errors anywhere else in the code base, because the functions were the last in the list/tuple anyway.

chrisseto: nice :+1:

scrapi/harvesters/biomed.py
fabianvf: yeah, I am not sure. I tried not to change the behavior of any of the harvesters

scrapi/harvesters/biomed.py
fabianvf: I don't know how to respond to that. I agree?

PR #131, opened 2015-03-27
Add until date to OAI harvesters

erinspace:


chrisseto:
Could you possibly write a test for this?

chrisseto:
lgtm.
Final thoughts @fabianvf ?

fabianvf:
Did you still intend to add it for the other harvesters in this PR, or will you do it separately @erinspace?

If separately, the :+1:

chrisseto:
It may not be relevant to other harvesters.

fabianvf:
most (all?) other providers support this feature. I think it would be better to have a unified way to interact with the harvesters.

erinspace:
most of the harvesters do support a range, but not all.

PR #141, opened 2015-04-13
Add fakejson harvester

walkerh:


fabianvf:
We have integrated this concept into our testing suite, to keep it separate from the production code base. Thank you for the suggestion!

PR #143, opened 2015-04-13
Add back flat file storage for testing and development

walkerh:
Code actually from Fabian.

https://gist.github.com/fabianvf/597f57ffe8351156bb98


fabianvf:
Awesome, that will be very helpful for local development.

PR #144, opened 2015-04-14
Add custom OAI schema for pubmed uris

erinspace:
# Purpose
Pubmed URIs all went to the same place, this adds a custom URI parser that finds the right url.

# Changes
Add new uri parser using existing helper function

# Side effects
Should only effect pubmed, old records will need to be renormalized!

### DIFF
--- a/scrapi/harvesters/pubmed.py
+++ b/scrapi/harvesters/pubmed.py
@@ -32,7 +32,14 @@ class PubMedHarvester(OAIHarvester):
         }
     }
 
-    schema = helpers.updated_schema(schemas.OAISCHEMA, updated)
+    schema = helpers.updated_schema(
+        schemas.OAISCHEMA,
+        {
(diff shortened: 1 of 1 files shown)


scrapi/harvesters/pubmed.py
```diff
@@ -5,14 +5,35 @@
...
     long_name = 'PubMed Central'
     url = 'http://www.ncbi.nlm.nih.gov/pmc/'
 
+    updated = {
+        "uris": {
+            "canonicalUri": ('//dc:identifier/node()', oai_extract_url_pubmed)
+        }
+    }
+
+    schema = helpers.updated_schema(schemas.OAISCHEMA, updated)
```
chrisseto: Can you merge these into one line?

```python
schema = helpers.updated_schema(schema, {
...
})
```

PR #149, opened 2015-04-14
Added harvester support for Stepic.org platform

mehanig:


scrapi/harvesters/stepic.py
```diff
@@ -0,0 +1,93 @@
...
+logger = logging.getLogger(__name__)
+
+
+def process_owner(owners_id):
+    logger.info(owners_id)
+    resp = requests.get("https://stepic.org:443/api/users/" + str(owners_id)).json()
+    try:
+        person = resp[u'users'][0]
+    except KeyError:
+        person = {u'first_name': 'None', u'last_name': 'None'}
```
fabianvf: Could you make these empty strings (```''```) instead? Otherwise, we will get contributors whose name displays as "None None", rather than "".

scrapi/harvesters/stepic.py
```diff
@@ -0,0 +1,93 @@
...
+        'email': '',
+        'sameAs': [],
+    }
+    return [owner]
+
+
+class StepicHarvester(JSONHarvester):
+    short_name = 'stepic'
+    long_name = 'Stepic.org Online Education Platform'
+    url = 'http://www.stepic.org/lesson'
```
fabianvf: This URL should correspond to a splash or info page for the service itself. In this case, I think ```https://www.stepic.org``` is the better choice. (These URLs are used to link to information about the sources we collect metadata from, so they are purely for human consumption).

scrapi/harvesters/stepic.py
```diff
@@ -0,0 +1,93 @@
...
+    count = 0
+
+    URL = 'https://stepic.org:443/api/lessons'
```
fabianvf: Is there a reason a port is specified here? And if so, will it remain consistent?

scrapi/harvesters/stepic.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def harvest(self, days_back=1):
+
+        search_url = self.URL
+        records = self.get_records(search_url)
+
+        record_list = []
+        for record in records:
+            doc_id = record['id']
+            logger.info(doc_id)
```
fabianvf: You can remove this logging statement.

scrapi/harvesters/stepic.py
```diff
@@ -0,0 +1,93 @@
...
+            )
+        return record_list
+
+    def get_records(self, search_url):
+        all_lessons = []
+        pk = 1
+        lesson = requests.get(search_url + "/" + str(pk))
+        resp = requests.get(self.URL + '?page=last').json()
+        last_lesson_id = resp['lessons'][-1]['id']
+        while pk < last_lesson_id:
```
fabianvf: Would this make more sense as a for loop?
```
for pk in xrange(1, last_lesson_id + 1):
    if lesson.status_code == 200:
        lesson_list = lesson.json()['lessons'][0]
        all_lessons.append(lesson_list)
    lesson = requests.get(search_url + "/" + str(pk))
```

fabianvf:
Just finished the code review, highlighted a few issues/questions I noticed/had. Other than that, looks good :+1: 

Thanks for sprinting with us, it has been really fun!

scrapi/harvesters/stepic.py
fabianvf: Could you remove this logging statement, or change it to logger.debug?

Soon as that change goes in, will merge :+1:

fabianvf: All right, I'm going to merge this in and get rid of the logger. Have fun in NYC!

PR #164, opened 2015-04-20
xml transformer no longer automatically tries to cast single-element lists to strings

fabianvf:
Ready for review. @erinspace @chrisseto


scrapi/base/helpers.py
```diff
@@ -2,11 +2,24 @@
...
 from copy import deepcopy
+import functools
+
 from nameparser import HumanName
 
 URL_REGEX = re.compile(ur'(https?://\S*\.\S*)')
 
 
+def single_result(l, default=''):
+    return l[0] if len(l) == 1 else default
```
fabianvf: Wasn't sure what the behavior should be if there was more than 1 result. Should it error, or just return the first element in the list?

fabianvf:
It turns out this PR is actually required for the otherProperties PR as well (found some errors caused by this bug). @chrisseto should I keep this separate or merge them into a single PR?

scrapi/base/schemas.py
```diff
@@ -7,18 +7,21 @@
...
 
 CONSTANT = lambda x: lambda *_, **__: x
 
+
 BASEXMLSCHEMA = {
-    "description": ('//dc:description/node()', lambda x: unicode(x.strip())),
-    "contributors": ('//dc:creator/node()', lambda x: default_name_parser(x.split(';'))),
-    "title": ('//dc:title/node()', lambda x: unicode(x.strip())),
-    "providerUpdatedDateTime": ('//dc:dateEntry/node()', lambda x: unicode(x.strip())),
+    "description": ('//dc:description/node()', compose(lambda x: unicode(x.strip()), single_result)),
```
chrisseto: Why not `compose(unicode, str.strip, single_result)`?

scrapi/base/transformer.py
```diff
@@ -72,8 +72,7 @@ class XMLTransformer(BaseTransformer):
     __metaclass__ = abc.ABCMeta
 
     def _transform_string(self, string, doc):
-        val = doc.xpath(string, namespaces=self.namespaces)
-        return unicode(val[0]) if len(val) == 1 else [unicode(v) for v in val] or ''
+        return doc.xpath(string, namespaces=self.namespaces)
```
chrisseto: Just to be clear, everything will now be a list?

fabianvf: Yes.

scrapi/base/schemas.py
fabianvf: That won't play nicely with unicode/python 3, will it? Also, is there actually any necessity that we cast to unicode anymore?

scrapi/base/transformer.py
fabianvf: There were several problems with the removed line, particularly if lists of lxml Elements were returned. I think it is better just to deal with the default xpath return values. I did add the single_result function to help if you expect just a single result.

fabianvf: Also, there are a variety of xpath commands that don't return lists, so if people wanted to use those it would be no problem.

scrapi/base/schemas.py
```diff
@@ -27,8 +30,8 @@
     "uris": {
         "canonicalUri": ('//dc:identifier/node()', oai_extract_url)
```
erinspace: Lots of times identifiers return multiple results, should we add the "single_result" here too?

fabianvf: I think ```oai_extract_url``` takes advantage of the multiple results to find a good URL.

erinspace:
This looks good to me, aside from my one identifiers question in the base OAI schema. 

Can you also add mention of this, particularly the helper single_result in the OSF SHARE documentation, too? Specifically in the [creating a harvester](https://osf.io/wur56/wiki/Creating%20a%20Harvester/) and [scrapi](https://osf.io/wur56/wiki/scrAPI/) sections.

scrapi/base/schemas.py
erinspace: cool!

scrapi/base/helpers.py
```diff
@@ -2,11 +2,24 @@
...
 from copy import deepcopy
+import functools
+
 from nameparser import HumanName
 
 URL_REGEX = re.compile(ur'(https?://\S*\.\S*)')
 
 
+def single_result(l, default=''):
+    return l[0] if len(l) >= 1 else default
```
chrisseto: `l[0] if l else default`

PR #168, opened 2015-04-22
Make it so that stepic can save non-200 results

fabianvf:
@erinspace @chrisseto


erinspace:
Looks good to me! :+1:

PR #176, opened 2015-04-23
otherProperties Updates

erinspace:
Also change tags to subjects, as per the schema

This will require a migration (just like pubmed) --Fabian


fabianvf:
Can you bring this up to date? I will then merge.

scrapi/base/helpers.py
```diff
@@ -1,14 +1,42 @@
...
+from copy import deepcopy
 
+from pycountry import languages
 from nameparser import HumanName
 
+
 URL_REGEX = re.compile(ur'(https?://\S*\.\S*)')
 
 
+def build_properties(*args):
```
chrisseto:
```python
[
    build_property(*args, **kwargs)
    for args, kwargs in 
    [
        pack(name, exp, desciption='foo'),
        pack(name, exp, uri='bar')
    ]
]
```

scrapi/base/helpers.py
```diff
@@ -98,3 +126,10 @@ def oai_process_contributors(*args):
 
 def pack(*args, **kwargs):
     return args, kwargs
+
+
+def language_code(language):
+    try:
+        return languages.get(name=language)
```
chrisseto: Does `.get` not have a defualt?

scrapi/base/transformer.py
```diff
@@ -28,18 +29,24 @@ def _transform(self, schema, doc):
...
+            elif isinstance(value, tuple):
+                transformed[key] = self._transform_tuple(value, doc)
             elif isinstance(value, basestring):
                 transformed[key] = self._transform_string(value, doc)
             elif callable(value):
                 transformed[key] = value(doc)
         return transformed
 
-    def _transform_iterable(self, l, doc):
+    def _transform_list(self, l, doc):
```
chrisseto: Does this really need to be a helper?

scrapi/base/helpers.py
fabianvf: no, it is pretty strange.

PR #188, opened 2015-05-01
More 0.6 fixes

fabianvf:
- Run provider map in tests.
- Ignore unicode-decode errors in copy_to_unicode
- Don't delete documents from cassandra (yet), do delete them from all elasticsearch indices.
- Delete old provider map before making a new one
- Add delete task

### DIFF
--- a/scrapi/util.py
+++ b/scrapi/util.py
@@ -1,4 +1,3 @@
-import six
 from datetime import datetime
 
 import pytz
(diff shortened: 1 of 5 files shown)


fabianvf:
@erinspace, ready for review

scripts/rename.py
```diff
@@ -35,8 +35,9 @@ def rename(source, target, dry=True):
             exceptions.append(e)
         else:
             if not dry:
-                doc.delete()
+                # doc.delete()
```
erinspace: So you still want this in there, just commented out?

scrapi/util.py
```diff
@@ -20,9 +19,9 @@ def copy_to_unicode(element):
         for idx, item in enumerate(element):
             element[idx] = copy_to_unicode(item)
     else:
-        element = ''.join(element)
         try:
-            element = six.u(element)
+            # A dirty way to convert to unicode in python 2 + 3.3+
```
chrisseto: (probably)

Issue #208, opened 2015-05-20
create invoke task for making tests for new harvesters


erinspace:

kms6bn:
@erinspace should we consider this closed?

erinspace:
@kms6bn I think this issue was originally referencing the automatic *creation* of a test for a new harvester, not running just one (which is AWESOME and totally needed). But I think that this issue still stands, we'd like to be able to auto create a new test w/yml file for a new harvester

PR #216, opened 2015-06-01
Added invoke test_harvesters to test all harvester tests

PatrickEGorman:


PatrickEGorman:
Related to https://github.com/fabianvf/scrapi/issues/208

fabianvf:
I think this has a similar functionality to the existing ```invoke test``` task. The harvester test generation referred to in #208 involves generating a request/response yaml file inside of the ```tests/vcr``` directory for a given harvester class.

PatrickEGorman:
Okay.  Regardless of how practical it is, I learned some about tasks and tests from it so no big loss.

tasks.py
```diff
@@ -88,6 +89,22 @@ def test(cov=True, verbose=False, debug=False):
...
+    """
+    cmd = 'py.test tests'
+    args = ''
+    if verbose:
+        args += ' -v'
+    if debug:
+        args += ' -s'
+    if cov:
+        args += ' --cov-report term-missing --cov-config .coveragerc --cov scrapi'
+    map(lambda x: run('py.test tests/' + x + args) if('harvester' in x) else(), os.listdir('tests'))
```
fabianvf: As an additional, stylistic note, I think maps are usually best used in functions where you don't have side-effects, but have a transformation that you want to perform on each element of the list. For example 
```
map(str, [1, 2, 3, 4])
```
Takes a list of integers and returns a list of strings. As another example, say you wanted only the first character of a list of strings, you could do:
```
map(lambda x: x[:1], ['hector', 'earl', 'lenny', 'larry', 'oswald'])
```
But when you have side-effects like the above function, and the return value of the function you are mapping isn't important to you, a for loop is probably the better choice.

fabianvf:
Yeah, absolutely. If you want me to give you a quick walkthrough of issue #208, I would be happy to. It's a pretty tricky little problem/

PatrickEGorman:
In a little I'd be glad.  I have to finish something up real quick

PR #217, opened 2015-06-02
add better document model for cassandra

fabianvf:
I think this is ready for review. @chrisseto @erinspace , any comments?

### DIFF
--- a/scrapi/processing/cassandra.py
+++ b/scrapi/processing/cassandra.py
@@ -49,7 +49,7 @@ class CassandraProcessor(BaseProcessor):
         self.send_to_database(**raw_doc.attributes).save()
 
     def send_to_database(self, docID, source, **kwargs):
-        documents = DocumentModel.objects(docID=docID, source=source)
+        documents = DocumentModelV2.objects(docID=docID, source=source)
         if documents:
             document = documents[0]
             if self.different(dict(document), dict(docID=docID, source=source, **kwargs)):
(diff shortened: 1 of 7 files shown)


scripts/util.py
```diff
@@ -4,27 +4,36 @@
...
+def ModelIteratorFactory(model, next_page):
+    def model_iterator(*sources):
+        q = model.objects.timeout(500).allow_filtering().all().limit(1000)
+        querysets = (q.filter(source=source) for source in sources) if sources else [q]
+        for query in querysets:
+            page = try_forever(list, query)
+            while len(page) > 0:
+                for doc in page:
+                    yield doc
+                page = try_forever(next_page, query, page)
```
erinspace: does this next_page need to be next_page_v1 or next_page_v2?

erinspace:
pass one done!

scripts/util.py
fabianvf: so if you notice the factory takes a model and a next page function. So the actual iterators, ```documents_v1``` or ```documents_v2``` are created with ```DocumentModel``` and ```next_page_v1``` or ```DocumentModelV2``` and ```next_page_v2```. So the original model will be with ```next_page_v1```, and the new model will be with ```next_page_v2```

erinspace: Got it! Cool!

scrapi/processing/cassandra.py
```diff
@@ -73,6 +73,48 @@ def different(self, old, new):
...
+
+    The schema contains denormalized raw document, denormalized
+    normalized (so sorry for the terminology clash) document, and
+    a list of version IDs that refer to previous versions of this
+    metadata.
+    '''
+    __table_name__ = 'documents_v2'
+
+    # Raw
+    source = columns.Text(primary_key=True, partition_key=True)
```
chrisseto: Does the ordering of these attributes matter?

scrapi/processing/cassandra.py
```diff
@@ -73,6 +73,48 @@ def different(self, old, new):
...
+class DocumentModelV2(models.Model):
+    '''
+    Defines the schema for a metadata document in cassandra
+
+    The schema contains denormalized raw document, denormalized
+    normalized (so sorry for the terminology clash) document, and
+    a list of version IDs that refer to previous versions of this
+    metadata.
+    '''
+    __table_name__ = 'documents_v2'
```
chrisseto: You may want to change the table name and possibly even the model name to something like documents correctly clustered/ordered. v2 sounds like the storage schema may have changed.
And later on if you decide to remove the DocumentModel and replace it with this one the name will make a bit more sense.

scripts/delete.py
```diff
@@ -1,6 +1,6 @@
 import logging
 
-from scripts.util import documents
+from scripts.util import documents_v2 as documents
```
chrisseto: Correct me if I'm wrong but `documents` is a generator and `documents_v2` is a migration script?
It doesn't look like they should be drop in replacements of each other

scripts/util.py
chrisseto: Why not `Mr.Meeseeks(next_page, query, page)`?

chrisseto:
In regards to the table naming it may be a bit cleaner to rename document model to DocumentModelOld rather than creating a new DocumentModelV2

chrisseto:
First pass finished :+1:

scripts/delete.py
fabianvf: this is actually a naming collision between the documents_v2 script and documents_v2 generator, which I should definitely resolve anyway.

scrapi/processing/cassandra.py
fabianvf: Not really, I just put source first because I thought it made sense. Documents are grouped first by source, and then within source ordered by docID

scripts/util.py
fabianvf: The behavior is actually subtly different. Where this just keeps trying the task, Mr.Meeseeks would continuously fork the task until it finally killed the server.

chrisseto:
Is there a migration script?

PR #220, opened 2015-06-08
Update migrations to use celery - rename task

erinspace:


tasks.py
```diff
@@ -36,12 +36,36 @@ def renormalize(sources=None):
 
 
 @task
-def rename(source, target, dry=True):
```
fabianvf: Can delete this task now

tasks.py
```diff
@@ -36,12 +36,36 @@ def renormalize(sources=None):
...
 
 @task
+def migrate(migration, kwargs_string, dry=True, async=False):
+    settings.CELERY_ALWAYS_EAGER = not async
+
+    from scrapi import migrations
+    from scrapi.tasks import migrate
+
+    kwargs = {
+        key.strip(): val.strip() for key, val in map(lambda x: x.split(':'), kwargs_string.split(','))
```
fabianvf: Maybe make a comment/docstring expaining the syntax of the kwargs string.

scrapi/util.py
```diff
@@ -36,3 +36,7 @@ def stamp_from_raw(raw_doc, **kwargs):
 
 def format_date_with_slashes(date):
     return date.strftime('%m/%d/%Y')
+
+
+def create_rename_iterable(documents, source, target, dry):
```
fabianvf: What is this for? If it is for a specific migration, I'd prefer it stay in the migrations file.

scrapi/migrations.py
```diff
@@ -0,0 +1,36 @@
+import logging
+# import functools
+
+from scrapi import tasks
+from scrapi import settings
+from scrapi.linter import RawDocument
+from scrapi.processing.elasticsearch import es
+from celery import Celery
+
+app = Celery()
```
fabianvf: I don't think you need this.

scrapi/migrations.py
```diff
@@ -0,0 +1,36 @@
...
+app.config_from_object(settings)
+
+logger = logging.getLogger()
+
+
+@tasks.task_autoretry(default_retry_delay=30, max_retries=5)
+def rename(doc, source=None, target=None, dry=True):
```
fabianvf: you may want to allow ```**kwargs``` here as well, just so that if you accidentally specified an extra option at the command line it wouldn't fail. Though we may want it to fail, just to make you be explicit.

scrapi/migrations.py
```diff
@@ -0,0 +1,36 @@
+import logging
+# import functools
```
fabianvf: Remove this.

scrapi/migrations.py
```diff
@@ -0,0 +1,36 @@
...
+
+from scrapi import tasks
+from scrapi import settings
+from scrapi.linter import RawDocument
+from scrapi.processing.elasticsearch import es
+from celery import Celery
```
fabianvf: Don't need this import either

scrapi/migrations.py
```diff
@@ -0,0 +1,36 @@
...
+
+logger = logging.getLogger()
+
+
+@tasks.task_autoretry(default_retry_delay=30, max_retries=5)
+def rename(doc, source=None, target=None, dry=True):
+    assert source and target, "To run this migration you need both a source and a target"
+    assert source != target, "Can't rename {} to {}, names are the same".format(source, target)
+
+    raw = RawDocument({
```
fabianvf: Rather than renormalizing it, we may be able to do something like
```
doc_attrs = dict(doc)
doc_attrs['source'] = target
DocumentModel.create(**attrs).save()
```

scrapi/tasks.py
```diff
@@ -35,7 +51,7 @@ def run_harvester(harvester_name, start_date=None, end_date=None):
     (start_harvest | normalization).apply_async()
 
 
-@app.task(default_retry_delay=300, max_retries=5)
+@app.task
```
fabianvf: Any reason these aren't retried?

scrapi/tasks.py
```diff
@@ -102,16 +118,30 @@ def normalize(raw_doc, harvester_name):
...
+        processing.process_normalized(raw_doc, normalized_doc, kwargs)
+    except Exception as e:
+        logger.info('Retrying withiin PROCESS NORMALIZED with exception {}'.format(e))
+        process_normalized.retry(exc=e)
+
+
+@app.task
+def migrate(migration, **kwargs):
+    count = 0
+    for doc in documents(kwargs['source']):
```
fabianvf: source won't be guaranteed for migrations, and I think you may have meant to put the rest of the function inside this for loop.

tasks.py
```diff
@@ -36,9 +36,22 @@ def renormalize(sources=None):
 
 
 @task
-def rename(source, target, dry=True):
-    from scripts.rename import rename
-    rename(source, target, dry)
+def migrate(migration, kwargs_string, dry=True, async=False):
+    settings.CELERY_ALWAYS_EAGER = not async
```
fabianvf: Is this necessary?

fabianvf:
1st pass done :+1:

scrapi/tasks.py
erinspace: Retrying was getting a bit complicated what with those tasks themselves using chains and other things - decided to move most of the retry attempts to another future PR

scrapi/migrations.py
```diff
@@ -0,0 +1,39 @@
...
+
+        es.delete(index=settings.ELASTIC_INDEX, doc_type=source, id=raw['docID'], ignore=[404])
+        es.delete(index='share_v1', doc_type=source, id=raw['docID'], ignore=[404])
+    logger.info('Deleted document from {} with id {}'.format(source, raw['docID']))
+
+
+def create_rename_iterable(documents, source, target, dry):
+    """ Will only be used if the migration is using async, and only if we
+    can get the chunks issue of retrying worked out."""
+    return [(doc, source, target, dry) for doc in documents]
```
fabianvf: Are you sure you want to do this? This could be a list of hundreds of thousands of documents, and milions in a few months.

scrapi/tasks.py
```diff
@@ -102,16 +118,30 @@ def normalize(raw_doc, harvester_name):
...
+    count = 0
+    for doc in documents(kwargs.get('source')):
+        count += 1
+
+        if kwargs.get('async'):
+            migration.s(doc, **kwargs).apply_async()
+        else:
+            migration(doc, **kwargs)
+
+    if kwargs.get('dry'):
```
fabianvf: This can probably be handled by the individual migration tasks, since some may not support dry runs.

erinspace: hard to move this dry run logging into the rename migration as it deals with each individual document, and so then would be logged as many times as there are documents!

scrapi/migrations.py
```diff
@@ -0,0 +1,99 @@
...
+from scrapi.processing.elasticsearch import es
+from scrapi.processing.cassandra import DocumentModel, DocumentModelOld
+
+
+logger = logging.getLogger()
+
+
+@tasks.task_autoretry(default_retry_delay=30, max_retries=5)
+def rename(doc, target=None, **kwargs):
+    assert target, "To run this migration you need both a source and a target"
```
fabianvf: Should update the message to just mention target

scrapi/processing/storage.py
```diff
@@ -20,8 +21,16 @@ def process_raw(self, raw):
         if not os.path.exists(os.path.dirname(filename)):
             os.makedirs(os.path.dirname(filename))
 
+        new_attrs = copy.deepcopy(raw.attributes)
+        if new_attrs.get('versions'):
+
+            if isinstance(new_attrs['versions'], list):
```
fabianvf: I think this can only be a list.

tests/test_processing.py
```diff
@@ -3,15 +3,27 @@
 
 from scrapi import settings
 
-settings.DEBUG = False
-settings.RAW_PROCESSING = ['storage', 'osf', 'foo', 'bar']
-settings.NORMALIZED_PROCESSING = ['storage', 'osf', 'foo', 'bar']
+# settings.DEBUG = False
```
fabianvf: Delete these

tests/utils.py
```diff
@@ -1,8 +1,9 @@
 from __future__ import unicode_literals
 
 from scrapi.base import XMLHarvester
+from scrapi.linter.document import RawDocument
```
fabianvf: △

PR #224, opened 2015-06-08
Update dryad to reject items not available

erinspace:


scrapi/harvesters/dryad.py
```diff
@@ -14,5 +19,17 @@ class DryadHarvester(OAIHarvester):
...
 
     base_url = 'http://www.datadryad.org/oai/request'
-    property_list = ['rights', 'format', 'relation', 'date', 'identifier', 'type']
+    property_list = ['rights', 'format', 'relation', 'date',
+                     'identifier', 'type', 'setSpec']
     timezone_granularity = True
+
+    def normalize(self, raw_doc):
+        str_result = raw_doc.get('doc')
+        result = etree.XML(str_result)
```
fabianvf: If ```raw_doc.get('doc')``` is None, this will still throw an error, and 'doc' is required, so you might as well just do ```result = etree.XML(raw_doc['doc'])```

scrapi/harvesters/dryad.py
```diff
@@ -14,5 +19,17 @@ class DryadHarvester(OAIHarvester):
...
+
+        status = (result.xpath('//dc:status/node()', namespaces=self.namespaces) or [''])[0]
+        if str(status).lower() == 'deleted' or str(status).lower() == 'item is not available':
```
fabianvf: you can do ```str(status).lower() in ['deleted', 'item is not available']```

PR #228, opened 2015-06-11
Update shareok harvester

erinspace:
to only harvest sets where all the sets are in the approved list


fabianvf:
This build failed, but only because we were unable to create a response for shareok (as their API is currently down)

PR #235, opened 2015-06-17
modified dataone.py to fit schema.

Johnetordoff:
**Purpose**
Put things stored in 'other properties' into proper places.
**Changes**
sorts urls, uris and contributor names into corrent catagory.
**Side Effects**
None that I know of.

### DIFF
--- a/scrapi/harvesters/dataone.py
+++ b/scrapi/harvesters/dataone.py
@@ -137,7 +137,8 @@ class DataOneHarvester(XMLHarvester):
             ('site', "arr[@name='site']/str/node()"),
             ('size', ("long[@name='size']/node()")),
             ('isDocumentedBy', "arr[@name='isDocumentedBy']/str/node()"),
-            ('serviceID', "str[@name='id']/node()")
+            ('serviceID', "str[@name='id']/node()"),
+            ('sku', "str[@name='sku']/node()")
         ),
         'freeToRead': {
             'startDate': ("bool[@name='isPublic']/node()", "date[@name='dateModified']/node()", lambda x, y: parse(y[0]).date().isoformat() if x else None)
@@ -145,13 +146,12 @@ class DataOneHarvester(XMLHarvester):
         'contributors': ("str[@name='author']/node()", "str[@name='submitter']/node()", "arr[@name='origin']/str/node()", "arr[@name='investigator']/str/node()", process_contributors),
         'uris': {
             'canonicalUri': ("str[@name='id']/node()", "//str[@name='dataUrl']/node()", lambda x, y: y[0] if 'http' in single_result(y) else x[0] if 'http' in single_result(x) else ''),
-            'objectUri': "str[@name='resourceMap']/node()".replace("doi:", "http://dx.doi.org/")
+            'objectUri': ("arr[@name='resourceMap']/str/node()", compose(lambda x: x.replace('doi:', 'http://dx.doi.org/'), single_result))
         },
         'tags': ("//arr[@name='keywords']/str/node()", lambda x: x if isinstance(x, list) else [x]),
         'providerUpdatedDateTime': ("str[@name='dateModified']/node()", compose(lambda x: parse(x).date().isoformat(), single_result)),
         'title': ("str[@name='title']/node()", single_result),
         'description': ("str[@name='abstract']/node()", single_result)
-
     }
 
     def harvest(self, start_date=None, end_date=None):
--- a/scrapi/harvesters/dataone.py
+++ b/scrapi/harvesters/dataone.py
@@ -30,7 +30,6 @@ logger = logging.getLogger(__name__)
 DEFAULT_ENCODING = 'UTF-8'
 DATAONE_SOLR_ENDPOINT = 'https://cn.dataone.org/cn/v1/query/solr/'
 
-
 def process_doi(service_id, doc_doi):
     doi_re = '10\\.\\d{4}/\\w*\\.\\w*(/\\w*)?'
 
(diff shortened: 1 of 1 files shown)


scrapi/harvesters/dataone.py
```diff
@@ -124,35 +126,32 @@ class DataOneHarvester(XMLHarvester):
...
+        'contributors': ("str[@name='author']/node()", "str[@name='submitter']/node()", "arr[@name='origin']/str/node()", "arr[@name='investigator']/str/node()", process_contributors),
         'uris': {
             'canonicalUri': ("str[@name='id']/node()", "//str[@name='dataUrl']/node()", lambda x, y: y[0] if 'http' in single_result(y) else x[0] if 'http' in single_result(x) else ''),
+            'objectUri': "str[@name='resourceMap']/node()".replace("doi:", "http://dx.doi.org/")
         },
         'tags': ("//arr[@name='keywords']/str/node()", lambda x: x if isinstance(x, list) else [x]),
         'providerUpdatedDateTime': ("str[@name='dateModified']/node()", compose(lambda x: parse(x).date().isoformat(), single_result)),
         'title': ("str[@name='title']/node()", single_result),
         'description': ("str[@name='abstract']/node()", single_result)
+
```
fabianvf: should remove this newline

scrapi/harvesters/dataone.py
```diff
@@ -124,35 +126,32 @@ class DataOneHarvester(XMLHarvester):
...
             ('serviceID', "str[@name='id']/node()")
         ),
         'freeToRead': {
             'startDate': ("bool[@name='isPublic']/node()", "date[@name='dateModified']/node()", lambda x, y: parse(y[0]).date().isoformat() if x else None)
         },
-        'contributors': ("str[@name='author']/node()", "str[@name='submitter']/node()", "arr[@name='origin']/str/node()", process_contributors),
+        'contributors': ("str[@name='author']/node()", "str[@name='submitter']/node()", "arr[@name='origin']/str/node()", "arr[@name='investigator']/str/node()", process_contributors),
         'uris': {
             'canonicalUri': ("str[@name='id']/node()", "//str[@name='dataUrl']/node()", lambda x, y: y[0] if 'http' in single_result(y) else x[0] if 'http' in single_result(x) else ''),
+            'objectUri': "str[@name='resourceMap']/node()".replace("doi:", "http://dx.doi.org/")
```
fabianvf: The format for scrAPI schemas is a little weird, but the first element here is an xpath command (```str[@name='resourceMap']/node()```). scrAPI will run this xpath command and return the results. However, if there is some post-processing that you want to do, you can turn the value into a tuple (```(elem1, elem2)```), and make the last element of the tuple a function. scrAPI will take the return values of all of the previous xpath commands (or functions, but usually xpath commads), and feed them into the last function as arguments.

So I think what you were trying to accomplish is achieved like this:
```python
...
{
...
    'objectUri': "str[@name='resourceMap']/node()", lambda x: x.replace('doi:', 'http://dx.doi.org/')
}
...
```

In case you haven't seen them before, lambdas are anonymous functions
```python
add1 = lambda x: x + 1
```
is equivalent to
```
def add1(x):
    return x + 1
```

They are very useful in situations like this, because we have many oneoff functions that are very simple, and we don't really need them to clutter the namespace.

scrapi/harvesters/dataone.py
```diff
@@ -124,35 +126,32 @@ class DataOneHarvester(XMLHarvester):
...
-            ('investigator', "arr[@name='investigator']/str/node()"),
-            ('origin', "arr[@name='origin']/str/node()"),
-            ('isPublic', ("bool[@name='isPublic']/node()")),
             ('readPermission', "arr[@name='readPermission']/str/node()"),
             ('replicaMN', "arr[@name='replicaMN']/str/node()"),
             ('replicaVerifiedDate', "arr[@name='replicaVerifiedDate']/date/node()"),
             ('replicationAllowed', ("bool[@name='replicationAllowed']/node()")),
             ('numberReplicas', ("int[@name='numberReplicas']/node()")),
             ('preferredReplicationMN', "arr[@name='preferredReplicationMN']/str/node()"),
-            ('resourceMap', "arr[@name='resourceMap']/str/node()"),
```
fabianvf: Where did you move this to?

scrapi/harvesters/dataone.py
```diff
@@ -124,35 +126,32 @@ class DataOneHarvester(XMLHarvester):
...
             ('rightsHolder', ("str[@name='rightsHolder']/node()")),
             ('scientificName', "arr[@name='scientificName']/str/node()"),
             ('site', "arr[@name='site']/str/node()"),
             ('size', ("long[@name='size']/node()")),
-            ('sku', ("str[@name='sku']/node()")),
```
fabianvf: Same as above

fabianvf:
1st pass done :+1:

scrapi/harvesters/dataone.py
Johnetordoff: it doesn't like  'objectUri': "str[@name='resourceMap']/node()", lambda x: x.replace('doi:', 'http://dx.doi.org/') because x is apparently a list object.

fabianvf: Yeah, I thought that would happen. An xpath command ending in ```/node()``` will return a list (maybe not always? I think always). We created the helper function ```single_result``` to fix this, and another helper function called ```compose``` to combine functions. So you should be able to do something like 
```python
objectUri': "str[@name='resourceMap']/node()", compose(lambda x: x.replace('doi:', 'http://dx.doi.org/', single_result))
```

Johnetordoff:
:+1: ? @fabianvf

fabianvf:
Looks much better. Have you run it locally?

Johnetordoff:
yes, it runs without error.

Johnetordoff:
@fabianvf  :+1: this time?

PR #244, opened 2015-06-20
Python 3 port

### LINKED ISSUE #171: Python 3 support
this would be great!

jamesturk:
all tests seem to pass on 2.7 and 3.4

this touches *a lot* so should definitely be eyed with tons of suspicion 

closes #171

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+sudo: false
 python:
     - "2.7"
     - "3.4"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
     - "2.7"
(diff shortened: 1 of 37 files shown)


fabianvf:
Let me know if you need any help/support getting the changes we discussed on Sunday in. Don't worry about the speed of validation for now, that is a separate issue.

jamesturk:
thanks. I'm on a trip and away from a laptop for a couple of weeks. I think
the main thing is replacement of http pretty in that one test with mock
like all the others

jamesturk:
I can get to it in a week or two but if someone else wants to that'd be
cool too

erinspace:
I'm on it! Thanks so much @jamesturk :)

PR #250, opened 2015-06-24
added new harvester using autooai. - Cyberleninka

Johnetordoff:
**Purpose**
Makes cyberlenkia searchable. Close issue #247 
**Changes**
uses autooai to add new harvester cyberlenkia.py
**Side Effect**
None that I know of.


fabianvf:
Will need a favicon and a test.

PR #252, opened 2015-06-25
Update base OAI harvester to not gather records recursively

erinspace:
For really really big record caches!!

### DIFF
--- a/scrapi/base/helpers.py
+++ b/scrapi/base/helpers.py
@@ -155,6 +155,12 @@ def language_code(language):
 
 
 def oai_get_records_and_token(url, throttle, force, namespaces):
+    """ Helper function to get the records and any resumptionToken
+    from an OAI request.
+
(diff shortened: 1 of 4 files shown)


scrapi/base/helpers.py
```diff
@@ -149,3 +152,21 @@ def language_code(language):
         return languages.get(name=language)
     except KeyError:
         return None
+
+
+def get_records_and_token(url, throttle, force, namespaces):
```
fabianvf: Might want to namespace this as OAI, since it will only work for oai. We could also move all the OAI stuff to its own file, which wouldn't be a terrible idea at this point... Imports would change from ```from scrapi.base import <x>``` to ```from scrapi.base.oai import <x>```. If you don't want to make this change now, understandable.

fabianvf: Also maybe add a docstring.

Issue #254, opened 2015-06-30
Make default null value None rather than '' in schemas


fabianvf:

fabianvf:
OR even better, just leave nulled fields blank...

erinspace:
Fixed in #321

PR #257, opened 2015-06-30
Allow non-required elements to be either a string or null

erinspace:
In the normalized schema. 

Starts to address #254 but the rest of it needs to happen in the harvesters


fabianvf:
With the data cleaning changes, I think this is no longer necessary.

PR #260, opened 2015-07-02
[WIP] Celery tasks in batches

erinspace:


scrapi/migrations.py
```diff
@@ -17,57 +17,62 @@
...
 
 
 @tasks.task_autoretry(default_retry_delay=30, max_retries=5)
-def delete(doc, sources=None, **kwargs):
-    assert sources, "To run this migration you need a source."
-    doc.timeout(5).delete()
-    es.delete(index=settings.ELASTIC_INDEX, doc_type=sources, id=doc.docID, ignore=[404])
-    es.delete(index='share_v1', doc_type=sources, id=doc.docID, ignore=[404])
+def delete(docs, sources=None, **kwargs):
+    for doc in docs:
```
fabianvf: If this task is retried, the first few docs will fail. If doc has already been deleted, we probably just want to skip it.

scrapi/migrations.py
```diff
@@ -17,57 +17,62 @@
...
+        assert sources, "To run this migration you need a source."
+        doc.timeout(5).delete()
```
fabianvf: If this task is retried, the first few docs will fail. If doc has already been deleted, we probably just want to skip it.

scrapi/tasks.py
```diff
@@ -3,6 +3,7 @@
 from datetime import date, timedelta
 
 from celery import Celery
+from itertools import islice
```
fabianvf: This is a builtin module.

scrapi/tasks.py
```diff
@@ -94,13 +95,13 @@ def spawn_tasks(raw, timestamps, harvester_name):
     process_raw.delay(raw)
 
 
-@app.task
+@task_autoretry(default_retry_delay=30, max_retries=5)
```
fabianvf: Do we want to make these numbers constants/settings?

scrapi/tasks.py
```diff
@@ -125,22 +126,23 @@ def process_normalized(normalized_doc, raw_doc, **kwargs):
...
-        else:
-            migration(doc, sources=sources, dry=dry, **kwargs)
+    docs = documents(*sources)
+    if async:
+        segment = list(islice(docs, group_size))
+        while segment:
+            migration.s(segment, sources=sources, dry=dry, **kwargs).apply_async()
+            segment = list(islice(docs, group_size))
+    else:
+        for doc in documents(*sources):
```
fabianvf: Should use the same docs created above.

PR #264, opened 2015-07-06
Add Harvester - Social Science Research Center

erinspace:
From the RSS feed


AndrewSallans:
Small correction to make in a few locations:  it's the "Social Science Research Network" (SSRN), not "Center".

erinspace:
Upgraded to hotfix in #265

PR #265, opened 2015-07-07
[hotfix] Add Social Science Research Network RSS feed

erinspace:
Upgrading #264 to a hotfox


erinspace:
Oops, this is against develop.

Continue following the chain from #264 along to #265 to #266

PR #268, opened 2015-07-07
[hotfix] Mock new sleep function to fix periodic test failing

erinspace:
Add and mock a new sleep function that should hopefully fix periodic test failures


fabianvf:
nicely done!

PR #269, opened 2015-07-07
[hotfix] update SHAREOK, add verify property

erinspace:
Update SHAREOK harvester to harvest more specific sets, add verify to base OAI harvester to deal with off ssl certs in some cases

### DIFF
--- a/scrapi/base/__init__.py
+++ b/scrapi/base/__init__.py
@@ -154,7 +154,7 @@ class OAIHarvester(XMLHarvester):
         records_url = self.base_url + self.RECORDS_URL
         request_url = records_url + self.META_PREFIX_DATE.format(start_date, end_date)
 
-        records = self.get_records(request_url, start_date, end_date, verify=self.verify)
+        records = self.get_records(request_url, start_date, end_date, self.verify)
 
         rawdoc_list = []
         for record in records:
--- a/scrapi/base/__init__.py
+++ b/scrapi/base/__init__.py
@@ -124,6 +124,7 @@ class OAIHarvester(XMLHarvester):
     approved_sets = None
     timezone_granularity = False
     property_list = ['date', 'type']
+    verify = True
 
     @property
     def schema(self):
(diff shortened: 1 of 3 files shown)


scrapi/base/__init__.py
```diff
@@ -169,8 +170,8 @@ def harvest(self, start_date=None, end_date=None):
 
         return rawdoc_list
 
-    def get_records(self, url, start_date, end_date, resump_token=''):
-        data = requests.get(url, throttle=self.timeout)
+    def get_records(self, url, start_date, end_date, verify, resump_token=''):
```
fabianvf: I think this function still has access to the class, so you can just use self.verify instead of passing it around.

scrapi/base/__init__.py
```diff
@@ -169,8 +170,8 @@ def harvest(self, start_date=None, end_date=None):
...
+        data = requests.get(url, throttle=self.timeout, verify=verify)
```
fabianvf: so this would be verify=self.verify

scrapi/base/__init__.py
```diff
@@ -186,7 +187,7 @@ def get_records(self, url, start_date, end_date, resump_token=''):
             base_url = url.replace(self.META_PREFIX_DATE.format(start_date, end_date), '')
             base_url = base_url.replace(self.RESUMPTION + resump_token, '')
             url = base_url + self.RESUMPTION + token[0]
-            records += self.get_records(url, start_date, end_date, resump_token=token[0])
+            records += self.get_records(url, start_date, end_date, resump_token=token[0], verify=verify)
```
fabianvf: This wouldn't need to be passed.

tests/test_oai_harvester.py
```diff
@@ -28,7 +29,7 @@ def harvest(self, start_date=None, end_date=None):
                                body=TEST_OAI_DOC,
                                content_type="application/XML")
 
-        records = self.get_records(request_url, start_date, end_date)
+        records = self.get_records(request_url, start_date, end_date, self.verify)
```
fabianvf: And you wouldn't need to pass it here either.

erinspace:
Good call! Knew there was a better way...

PR #270, opened 2015-07-08
[hotfix] Update OSF harvester to honor period given to retract public registrations

erinspace:
Harvests will now automatically happen for a 2 day period starting 4 days in the past, to honor the 48 hours promised to contributors to public projects to make the registration private.

### DIFF
--- a/scrapi/harvesters/osf.py
+++ b/scrapi/harvesters/osf.py
@@ -16,7 +16,6 @@ from datetime import date, timedelta
 from nameparser import HumanName
 
 from scrapi import requests
-from scrapi import settings
 from scrapi.base import JSONHarvester
 from scrapi.linter.document import RawDocument
 from scrapi.base.helpers import build_properties
--- a/scrapi/harvesters/osf.py
+++ b/scrapi/harvesters/osf.py
@@ -96,9 +96,10 @@ class OSFHarvester(JSONHarvester):
         }
 
     def harvest(self, start_date=None, end_date=None):
-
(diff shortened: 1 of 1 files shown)


fabianvf:
./scrapi/harvesters/osf.py:19:1: F401 'settings' imported but unused

scrapi/harvesters/osf.py
```diff
@@ -96,9 +95,10 @@ def schema(self):
         }
 
     def harvest(self, start_date=None, end_date=None):
-
-        start_date = start_date or date.today() - timedelta(settings.DAYS_BACK)
```
fabianvf: This meas that we can't harvest back a configurable number of days for the OSF.

scrapi/harvesters/osf.py
```diff
@@ -96,9 +95,10 @@ def schema(self):
...
-        end_date = end_date or date.today()
+        # Always harvest a 2 day period starting 2 days back to honor time given
```
fabianvf: Why do we always harvest a 2 day period?

scrapi/harvesters/osf.py
erinspace: We can if we need to! If start date is passed in that will still be used, this just defaults to 4 days back

scrapi/harvesters/osf.py
erinspace: I just added that in as that seemed to be the standard in other harvesters - 2 days just to make sure we cover everything in the gaps between harvests

PR #271, opened 2015-07-08
[Ready for Review] Simple postgres processor and django rest API for web access

erinspace:


api/api/settings.py
```diff
@@ -0,0 +1,129 @@
...
+https://docs.djangoproject.com/en/1.8/ref/settings/
+"""
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+import os
+
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+# Quick-start development settings - unsuitable for production
```
fabianvf: Just noting that we should come back to this pre-merge

api/api/settings.py
```diff
@@ -0,0 +1,129 @@
...
+
+USE_L10N = True
+
+USE_TZ = True
+
+
+# Static files (CSS, JavaScript, Images)
+# https://docs.djangoproject.com/en/1.8/howto/static-files/
+
+STATIC_URL = '/static/'.format(DOMAIN)
```
fabianvf: the format does nothing here.

tests/test_migrations.py
```diff
@@ -11,14 +11,15 @@
...
-from scrapi.migrations import document_v2_migration
 
 # Need to force cassandra to ignore set keyspace
 from scrapi.processing.cassandra import CassandraProcessor, DocumentModel
+from scrapi.processing.postgres import PostgresProcessor
 
 from . import utils
 
 test_cass = CassandraProcessor()
+test_postgress = PostgresProcessor()
```
fabianvf: postgres has 1 s

erinspace:
Ok. This is a python path and imports issue. @fabianvf, halp!

fabianvf:
There are a few more things we need to think about. Cassandra is also responsible for request caching, so we need to either decide to use cassandra and postgres side-by-side for a while, or to port all that over to postgres. The migrations also all currently run on cassandra only, so we'd need to figure out what we want to do about that (keep cassandra migrations, make an interface for iterating over an arbitrary database, move all migrations to postgres, etc). And of course, we also have to worry about postgres deployment.

Personally, I am fine with incremental improvements, and keeping request caching in cassandra for now and addressing that in a later PR (maybe moving to just using file storage/memory, like the sunlight foundation did).

api/api/settings/defaults.py
```diff
@@ -0,0 +1,131 @@
...
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+import os
+
+from django.conf import global_settings
+
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = '...'
```
fabianvf: Do we need this? What is it for? Should it be a setting that is not version controlled?

api/api/settings/defaults.py
```diff
@@ -0,0 +1,131 @@
...
+For the full list of settings and their values, see
+https://docs.djangoproject.com/en/1.7/ref/settings/
+"""
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+import os
+
+from django.conf import global_settings
+
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
```
fabianvf: why are the ```os.path.dirname```s chained together like this?

api/api/settings/defaults.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = '...'
+
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG = False
+
+TEMPLATE_DEBUG = True
```
fabianvf: Should this be on by default?

api/api/settings/defaults.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+# Database
+# https://docs.djangoproject.com/en/1.7/ref/settings/#databases
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+        'NAME': 'scrapi'
+        # 'USER': 'name',
```
fabianvf: Do we need any of this commented out stuff?

api/webview/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+from django_pgjson.fields import JsonField
+
+
+class Document(models.Model):
+    source = models.CharField(max_length=100)
+    docID = models.CharField(max_length=100)
+
+    providerUpdatedDateTime = models.DateTimeField(null=True)
```
fabianvf: why the ```null=True```?

erinspace: Because the way we have the "approved sets" currently working, documents are all processed as raw documents, and then only the ones that are approved are then normalized. We did that in case anyone changed the sets they wanted to approve, so we wouldn't have to scrape any new material. Since the providerUpdatedDateTime is added during normalization, it's never added to docs that are just raw.

scrapi/processing/postgres.py
```diff
@@ -0,0 +1,47 @@
...
+    NAME = 'postgres'
+
+    @events.logged(events.PROCESSING, 'raw.postgres')
+    def process_raw(self, raw_doc):
+        source, docID = raw_doc['source'], raw_doc['docID']
+        document = self._get_by_source_id(Document, source, docID) or Document(source=source, docID=docID)
+
+        modified_doc = copy.deepcopy(raw_doc.attributes)
+        if modified_doc.get('versions'):
+            modified_doc['versions'] = str(modified_doc['versions'])
```
fabianvf: Why are you casting a list to a string here? You may want to do ```map(str, modified_doc['versions'])``` instead

tests/conftest.py
```diff
@@ -13,6 +13,7 @@
 settings.CELERY_EAGER_PROPAGATES_EXCEPTIONS = True
 database._manager.keyspace = 'test'
 
+
```
fabianvf: Is this newline necessary?

api/api/settings/defaults.py
erinspace: We do need this! This is just a placeholder, the real one goes in local.py. This particular settings config is directly from the SHARE Registration django setup done by Michael, so we can ask him why this particular thing

scrapi/processing/postgres.py
erinspace: good call!

api/webview/models.py
fabianvf: Cool, makes sense.

api/api/settings/defaults.py
erinspace: Michael? We can double check with him when we get it ready for prod!

api/api/settings/defaults.py
erinspace: Same -- we'll double check settings stuff!

fabianvf:
Can you update the README as well?

api/webview/tests.py
```diff
@@ -0,0 +1,29 @@
...
+import django
+from django.test import TestCase
+from rest_framework.test import APIRequestFactory
+
+from api.webview.views import DocumentList
+
+django.setup()
+
+
+# TODO - make this work without Django.
```
fabianvf: ?

requirements.txt
```diff
@@ -16,5 +16,10 @@ furl==0.4.4
 jsonschema==2.4.0
 jsonpointer==1.7
 pycountry==1.10
+djangorestframework==3.1.3
+Django==1.8.2
+django-pgjson==0.3.1
+django-cors-headers==1.1.0
+psycopg2==2.6.1
```
fabianvf: shouldn't pytest-django be in here as well?

scrapi/migrations.py
```diff
@@ -36,10 +39,58 @@ def rename(docs, target=None, **kwargs):
...
+            'timestamps': doc.timestamps,
+            'versions': doc.versions
+        }, validate=False)
+
+        normed = util.doc_to_normed_dict(doc)
+
+        # Create the normalized, don't validate b/c its been done once already
+        normalized = NormalizedDocument(normed, validate=False)
+
+        if target_db == 'postgres':
```
fabianvf: why not use the processing function ```get_processor(processor_name)```?
Plus that way you don't need to import them up top.

scrapi/migrations.py
```diff
@@ -100,6 +151,7 @@ def next_page_source_partition(query, page):
 
 documents_old = ModelIteratorFactory(DocumentModelOld, next_page_old)
 documents = ModelIteratorFactory(DocumentModel, next_page_source_partition, default_args=registry.keys())
+postgres_documents = Document.objects.all()
```
fabianvf: I would make it so that each processor exposes an iterator, rather than adding more of these

scrapi/processing/postgres.py
```diff
@@ -0,0 +1,134 @@
...
+from api.webview.models import HarvesterResponse, Document
+
+from scrapi import events
+from scrapi.processing.base import BaseProcessor, BaseHarvesterResponse, BaseDatabaseManager
+
+
+logger = logging.getLogger(__name__)
+
+
+class DatabaseManager(BaseDatabaseManager):
```
fabianvf: Erin, can you give this a look over and make sure it is correct?

scrapi/util.py
```diff
@@ -51,3 +55,32 @@ def json_without_bytes(jobj):
         if isinstance(v, six.binary_type):
             jobj[k] = v.decode('utf8')
     return jobj
+
+
+def doc_to_normed_dict(doc):
```
fabianvf: should this be part of the migrations?  I don't see it being useful outside of that scenario. Unless I missed something?

README.md
```diff
@@ -16,6 +16,7 @@ scrapi
     - Install requirements
     - Install Elasticsearch
```
fabianvf: Did you add the cp `api/api/settings/local-dist.py api/api/settings/local.py` to the README?

api/api/settings/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .defaults import *
+
+try:
+    from .local import *
+except ImportError as error:
+    raise ImportError("No local.py settings file found.")
```
fabianvf: Can also give them the `cp` command to fix it

api/api/settings/defaults.py
fabianvf: Oh, I see, it is trying to get the project path I think. Would this not also be covered by just getting the working directory?

api/webview/views.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def perform_create(self, serializer):
+        serializer.save(source=self.request.user)
+
+    def get_queryset(self):
+        """ Return all documents
+        """
+        queryset = Document.objects.all()
+
+        return queryset
```
fabianvf: why not just return `Document.objects.all()`?

api/webview/views.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def perform_create(self, serializer):
+        serializer.save(source=self.request.user)
+
+    def get_queryset(self):
+        """ Return queryset based on source
+        """
+        queryset = Document.objects.filter(source=self.kwargs['source'])
+
+        return queryset
```
fabianvf: Same as above

api/webview/views.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Retrieve one particular document.
+    """
+    try:
+        all_sources = Document.objects.filter(source=source)
+        document = all_sources.get(docID=docID)
+    except Document.DoesNotExist:
+        return Response(status=404)
+
+    if request.method == 'GET':
```
fabianvf: can the request method ever not be get?

requirements.txt
fabianvf: nvm, is in dev reqs

scrapi/base/__init__.py
```diff
@@ -89,7 +89,8 @@ class JSONHarvester(BaseHarvester, JSONTransformer):
     def normalize(self, raw_doc):
         transformed = self.transform(json.loads(raw_doc['doc']), fail=settings.RAISE_IN_TRANSFORMER)
         transformed['shareProperties'] = {
-            'source': self.short_name
+            'source': self.short_name,
+            'docID': raw_doc['docID']
```
fabianvf: do we want to stick filetype in here as well, just in case anyone ever wants to parse it out?

scrapi/base/__init__.py
```diff
@@ -100,7 +101,8 @@ class XMLHarvester(BaseHarvester, XMLTransformer):
     def normalize(self, raw_doc):
         transformed = self.transform(etree.XML(raw_doc['doc']), fail=settings.RAISE_IN_TRANSFORMER)
         transformed['shareProperties'] = {
-            'source': self.short_name
+            'source': self.short_name,
+            'docID': raw_doc['docID']
```
fabianvf: same as above

scrapi/migrations.py
```diff
@@ -36,10 +39,58 @@ def rename(docs, target=None, **kwargs):
...
+    TODO - make this source_db agnostic. Should happen along with larger migration refactor
+    """
+    assert target_db, 'Please specify a target db for the migration -- either postgres or elasticsearch'
+    assert target_db in ['postgres', 'elasticsearch'], 'Invalid target database - please specify either postgres or elasticsearch'
+
+    for doc in docs:
+
+        if not doc.doc:
+            # corrupted database item has no doc element
+            logger.info('Could not migrate document from {} with id {}'.format(doc.source, doc.docID))
```
fabianvf: actually. may want to send this to the sentry logger, so that we can see how many we've missed.

scrapi/migrations.py
fabianvf: Which is where the idea of the CANONICAL_PROCESSOR came from, actually. That way when doing migrations we'd read out of the correct database.

scrapi/processing/postgres.py
```diff
@@ -0,0 +1,134 @@
...
+            return None
+
+
+class HarvesterResponseModel(BaseHarvesterResponse):
+
+    response = None
+
+    def __init__(self, *args, **kwargs):
+        if kwargs:
+            self.response = HarvesterResponse(key=kwargs['method'].lower() + kwargs['url'].lower(), *args, **kwargs)
```
fabianvf: This is really long

fabianvf: Think it was me though

api/webview/models.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class Document(models.Model):
+    source = models.CharField(max_length=255)
+    docID = models.TextField()
+
+    providerUpdatedDateTime = models.DateTimeField(null=True)
+
+    raw = JsonField()
+    normalized = JsonField()
```
fabianvf: Can this be null?

README.md
```diff
@@ -1,24 +1,25 @@
 scrapi
 ======
 
-```master``` build status: [![Build Status](https://travis-ci.org/CenterForOpenScience/scrapi.svg?branch=master)](https://travis-ci.org/CenterForOpenScience/scrapi)
+```master``` build status: [![Build Status](https://travis-ci.org/fabianvf/scrapi.svg?branch=master)](https://travis-ci.org/fabianvf/scrapi)
```
fabianvf: Not sure what happened here, must have been a bad merge. These should definitely point at CenterForOpenScience/scrapi

scrapi/migrations.py
```diff
@@ -19,97 +14,87 @@
...
-            'filetype': doc.filetype,
-            'timestamps': doc.timestamps,
-            'versions': doc.versions
-        })
+        raw_doc = doc.raw
+        new_doc = copy.copy(raw_doc.attributes)
+        new_doc['source'] = target
 
-        assert doc.source != target, "Can't rename {} to {}, names are the same.".format(doc.source, target)
+        raw = RawDocument(new_doc, validate=False)
```
fabianvf: May want to have the option to run migrations with validate/clean, could make it another flag. Can be in a separate PR though, let's just get this merged for now.

scrapi/migrations.py
```diff
@@ -19,97 +14,87 @@
...
+@tasks.task_autoretry(default_retry_delay=30, max_retries=5)
+def cross_db(docs, target_db=None, index=None, **kwargs):
+    """
+    Migration to go between
+        cassandra > postgres
+        postgres > cassandra
+        cassandra > elasticsearch
+        postgres > elasticsearch
+
+    source db will be set to the CANONICAL_PROCESSOR specified in settings
```
fabianvf: I would make source db _default_ to CANONICAL_PROCESSOR, but still have the option to pass it in.

scrapi/migrations.py
erinspace: Agreed! I'll make a jira improvement ticket for it so we remember

scrapi/migrations.py
```diff
@@ -19,97 +14,87 @@
...
     for doc in docs:
-        raw = RawDocument({
-            'doc': doc.doc,
-            'docID': doc.docID,
-            'source': target,
-            'filetype': doc.filetype,
-            'timestamps': doc.timestamps,
-            'versions': doc.versions
-        })
+        raw_doc = doc.raw
```
fabianvf: `new_doc = copy.copy(raw_doc.attributes)`? Also might want to do a `deepcopy` here.

scrapi/migrations.py
```diff
@@ -19,97 +14,87 @@
...
+        cassandra > elasticsearch
+        postgres > elasticsearch
+
+    source db can be passed in to the migrate task, and will default to the CANONICAL_PROCESSOR specified in settings
+    target_db will be specified when the task is called
+    """
+    assert target_db, 'Please specify a target db for the migration -- either postgres or elasticsearch'
+    assert target_db in ['postgres', 'cassandra', 'elasticsearch'], 'Invalid target database - please specify either postgres, cassandra or elasticsearch'
+    for doc in docs:
+        if not doc.raw.attributes['doc']:
```
fabianvf: I think `doc.raw['doc']` should work. We should probably avoid accessing the internals of `Raw/NormalizedDocument`s where available (structure could easily change in the future...)

scrapi/migrations.py
```diff
@@ -19,97 +14,87 @@
...
+            # corrupted database item has no doc element
+            message = 'Could not migrate document from {} with id {}'.format(doc.raw.attributes['source'], doc.raw.attributes['docID'])
```
fabianvf: Could you make this message more specific to the error?

scrapi/migrations.py
```diff
@@ -19,97 +14,87 @@
...
+            log_to_sentry(message)
+            logger.info(message)
+            continue
 
-            es.delete(index=settings.ELASTIC_INDEX, doc_type=doc.source, id=raw['docID'], ignore=[404])
-            es.delete(index='share_v1', doc_type=doc.source, id=raw['docID'], ignore=[404])
+        raw = doc.raw
```
fabianvf: `raw, normalized = doc.raw, doc.normalized`, if you want, just to make it a little less noisy. Though is there any reason to break it out of the tuple?

scrapi/migrations.py
```diff
@@ -19,97 +14,87 @@
...
 
-        logger.info('Deleted document from {} with id {}'.format(doc.source, raw['docID']))
+        target_processor = get_processor(target_db)
+
+        if not kwargs.get('dry'):
+            target_processor.process_raw(raw)
+
+            try:
+                if target_db == 'elasticsearch':
+                    target_processor.process_normalized(raw, normalized, index=index)
```
fabianvf: Should be able to rely on the default index here, this `if` isn't necessary.

scrapi/migrations.py
```diff
@@ -19,97 +14,87 @@
...
     for doc in docs:
-        raw = RawDocument({
-            'doc': doc.doc,
-            'docID': doc.docID,
-            'source': doc.source,
-            'filetype': doc.filetype,
-            'timestamps': doc.timestamps,
-            'versions': doc.versions
-        })
+        raw_doc = doc.raw
```
fabianvf: Same question as above.

README.md
erinspace: yes!

tests/utils.py
```diff
@@ -21,6 +21,18 @@
     'source': 'test'
 }
 
+POSTGRES_RAW_DOC = {
```
fabianvf: Is there any way this could be merged with the real raw_doc? If not that's fine, just curious

tests/test_migrations.py
```diff
@@ -32,76 +28,190 @@ def harvester():
...
+
+    if destination != 'elasticsearch':
+        scrapi.processing.elasticsearch.es = real_es
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('destination', ['postgres', 'cassandra'])
+def test_no_doc_element_cross_db(destination, monkeypatch, index='test'):
+
+    if scrapi.settings.CANONICAL_PROCESSOR == destination:
```
fabianvf: Why is this case skipped?

tests/test_migrations.py
```diff
@@ -32,76 +28,190 @@ def harvester():
...
-        RAW['doc'] = str(RAW['doc'])
-    DocumentModelOld.create(**RAW.attributes).save()
-    queryset = DocumentModel.objects(docID=RAW['docID'], source=RAW['source'])
-    assert len(queryset) == 0
-    tasks.migrate_to_source_partition(dry=False)
-    queryset = DocumentModel.objects(docID=RAW['docID'], source=RAW['source'])
-    assert len(queryset) == 1
+@pytest.mark.elasticsearch
+@pytest.mark.parametrize('canonical', ['postgres', 'cassandra'])
+@pytest.mark.parametrize('destination', ['postgres', 'cassandra', 'elasticsearch'])
```
fabianvf: This is awesome parametrization.

tests/test_migrations.py
erinspace: Using the cross_db migration for a database to itself didn't make sense to me, should it not be skipped? Like, cross_db postgres to postgres?

This just uses the canonical processor as the source, but I could cycle through everything (just didn't want to add more load)

tests/utils.py
erinspace: This was done just cause the doc item had to be different (bytes vs not) but I can definitely make this less verbose

scrapi/processing/storage.py
```diff
@@ -35,3 +35,6 @@ def write(self, source, doc_id, filename, content):
 
         with open(filepath, 'w') as f:
             f.write(json.dumps(json_without_bytes(content), indent=4))
+
+    def documents(self, *sources):
+        pass
```
fabianvf: maybe `raise NotImplementedError` instead?

scrapi/processing/elasticsearch.py
```diff
@@ -89,6 +119,18 @@ def process_normalized_v1(self, raw_doc, normalized, date):
             id=raw_doc['docID']
         )
 
+    def get(self, index, source):
+        results = self.manager.es.search(index=index, doc_type=source)
```
fabianvf: There is actually an `es.get` method

scrapi/processing/elasticsearch.py
```diff
@@ -20,37 +22,65 @@
...
 
 
 class ElasticsearchProcessor(BaseProcessor):
     NAME = 'elasticsearch'
 
-    def process_normalized(self, raw_doc, normalized, index=settings.ELASTIC_INDEX):
+    manager = DatabaseManager()
+
+    def documents(self, *sources):
+        pass
```
fabianvf: might want to just `raise NotImplementedError`. Fail fast, fail hard

scrapi/processing/cassandra.py
```diff
@@ -77,8 +160,85 @@ def different(self, old, new):
...
+                for doc in page:
+                    yield DocumentTuple(self.to_raw(doc), self.to_normalized(doc))
+                page = try_n_times(5, self.next_page, query, page)
+
+    def next_page(self, query, page):
+        return list(query.filter(docID__gt=page[-1].docID))
+
+    def to_raw(self, doc):
+
+        doc_items = dict(doc)
```
fabianvf: Probably don't need to convert it to a dict, I think all the items should be reachable with just
```python
doc.doc
doc.docID
doc.source
doc.filetype
```
etc

scrapi/processing/cassandra.py
```diff
@@ -77,8 +160,85 @@ def different(self, old, new):
...
+        # make the new dict actually contain real items
+        normed = {}
+        for key, value in dict(doc).items():
+            try:
+                normed[key] = json.loads(value)
+            except (ValueError, TypeError):
+                normed[key] = value
+
+        # Remove empty values and strip down to only normalized fields
+        do_not_include = ['docID', 'doc', 'filetype', 'timestamps', 'source']
```
fabianvf: You could move this up to your first iteration through the dictionary, and just not add them, rather than adding them and then deleting them.

scrapi/processing/cassandra.py
```diff
@@ -77,8 +160,85 @@ def different(self, old, new):
...
+        for key in list(normed.keys()):
+            if not normed[key]:
+                del normed[key]
+
+        for key in list(normed.keys()):
```
fabianvf: see above.

scrapi/processing/cassandra.py
```diff
@@ -77,8 +160,85 @@ def different(self, old, new):
...
+            if key in do_not_include:
+                del normed[key]
+
+        if normed.get('versions'):
+            normed['versions'] = list(map(str, normed['versions']))
+
+        # No datetime means the document wasn't normalized (probably wasn't on the approved list)
+        if normed.get('providerUpdatedDateTime'):
+            normed['providerUpdatedDateTime'] = normed['providerUpdatedDateTime'].isoformat()
```
fabianvf: Might want to use the `scrapi.base.helpers.datetime_formatter` function instead, that way we can be sure it'll be in the expected form.

scrapi/processing/cassandra.py
```diff
@@ -77,8 +160,85 @@ def different(self, old, new):
...
+            'filetype': doc_items['filetype'],
+            'timestamps': doc_items['timestamps'],
+            'versions': doc_items['versions']
+        }, validate=False, clean=False)
+
+    def to_normalized(self, doc):
+        # make the new dict actually contain real items
+        normed = {}
+        for key, value in dict(doc).items():
+            try:
```
fabianvf: `if value and key not in do_not_include:`

fabianvf:
:+1:

PR #276, opened 2015-07-10
Fix language harvesting

fabianvf:


scrapi/base/helpers.py
```diff
@@ -144,8 +144,16 @@ def pack(*args, **kwargs):
...
+
+    e.g. ['English', 'Russian'] -> ['eng', 'rus']
+    '''
+    return filter(lambda x: x, map(get_code, langs))
+
+
+def get_code(language):
     try:
-        return languages.get(name=language)
+        return languages.get(name=language).bibliographic
```
erinspace: this key error is confusing - does .bibliographic access a list of some kind?

fabianvf: no, but the ```.get(name=language)``` will throw a key error if it is not found.

erinspace: coolcoolcool! Assumed languages was a dict, but I get it now!!

fabianvf:
ping @erinspace

PR #293, opened 2015-07-24
!! Figuring out postgres travis errors

erinspace:


erinspace:
made another testing PR as this one got too out of hand

PR #304, opened 2015-07-28
Add raising a value error if no uri was returned for cannonical uri

erinspace:
For OAI harvesters - a ValueError will be raised if no URI is returned for a particular record

### DIFF
--- a/scrapi/base/helpers.py
+++ b/scrapi/base/helpers.py
@@ -131,6 +131,7 @@ def oai_extract_url(identifiers):
                 return found_url
         except AttributeError:
             continue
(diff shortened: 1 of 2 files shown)


scrapi/base/helpers.py
```diff
@@ -131,6 +131,7 @@ def oai_extract_url(identifiers):
                 return found_url
         except AttributeError:
             continue
+    raise ValueError('No Cannonical URI was returned for this record.')
```
fabianvf: Canonical

fabianvf:
:+1: Will merge when build passes

PR #309, opened 2015-07-30
[WIP] Add uri logging to postgres normalized document

erinspace:
## Purpose
SHARE collects a lot of URIs, and it would be nice to also collect information about where those uris actually go. This is the begnings o f a processor that will follow those links and save information about their status, and where they resolve. 

## TODO
There is a lot to be done in order for this plan to be viable, most notably creating a central "control" system that keeps track of how many times we've hit a particular website with requests according to their terms of service.


fabianvf:
I would make this a new celery task and just make it an additional part of the pipe, I think.

erinspace:
Moving development to #389 for some more generalized post processing of both URLs and Contributors together, adding a new postgres model instead of just sticking it in the normalized document

PR #310, opened 2015-07-30
Add ScholarsArchive@OSU harvester

### LINKED ISSUE #294: New Harvester - ScholarsArchive@OSU
OAI base url - http://ir.library.oregonstate.edu/oai/request
Use autooai - https://github.com/erinspace/autooai

erinspace:
Also add new helper function for extracting a DOI from a block of text, and test for that function.

Closes #294


scrapi/harvesters/scholarsarchiveosu.py
```diff
@@ -0,0 +1,29 @@
...
+from __future__ import unicode_literals
+
+from scrapi.base import helpers
+from scrapi.base import OAIHarvester
+
+
+class ScholarsarchiveosuHarvester(OAIHarvester):
+    short_name = 'scholarsarchiveosu'
+    long_name = 'ScholarsArchive@OSU'
+    url = 'http://ir.library.oregonstate.edu/oai/request'
```
fabianvf: I don't think this is the correct url for the service (should point to a landing page or something)

PR #316, opened 2015-08-03
Update many OAI URLs to point to real source pages

erinspace:
## Purpose 
The Auto-OAI harvest creator was not saving correct urls that are saved to the provider map - instead, was using the base OAI urls that usually don't direct anywhere useful.

This PR updates those that I found to be actual home pages for sources.

## Changes
One line in a bunch of different harvesters.  This really should only affect the SHARE OSF search page - clicking on the sources in each search result - along with the SHARE registration page that has the sources listed.


fabianvf:
:+1:

PR #321, opened 2015-08-10
Feature/clean data on validation

fabianvf:
Makes validation much stricter. Datetimes must have, timezone information, email addresses cannot be empty strings. Also adds an optional step before validation, where falsey leaf values that are not required are stripped out of the document. This allows you to default to null values in non-required field without getting the null value validation errors. Also update all harvesters to conform. Ready for review, @erinspace 

P.S. Also stripped whitespace off the normalized.json, just couldn't take it anymore.

### DIFF
--- a/scrapi/harvesters/bhl.py
+++ b/scrapi/harvesters/bhl.py
@@ -12,7 +12,6 @@ def institution_name_parser(names):
     for inst in names:
         contributor = {
             'name': inst,
-            'email': '',
             'sameAs': []
         }
         contributor_list.append(contributor)
--- a/scrapi/harvesters/crossref.py
+++ b/scrapi/harvesters/crossref.py
@@ -13,7 +13,6 @@ from datetime import date, timedelta
 
 from six.moves import xrange
 from nameparser import HumanName
-from dateutil.parser import parse
 
 from scrapi import requests
 from scrapi import settings
--- a/scrapi/harvesters/crossref.py
+++ b/scrapi/harvesters/crossref.py
@@ -30,10 +30,8 @@ def process_contributor(author, orcid):
         'givenName': name.first,
         'additionalName': name.middle,
         'familyName': name.last,
-        'sameAs': []
+        'sameAs': [orcid] if orcid else []
     }
-    if orcid:
-        ret['sameAs'].append(orcid)
     return ret
 
 
--- a/scrapi/harvesters/crossref.py
+++ b/scrapi/harvesters/crossref.py
@@ -19,21 +19,23 @@ from scrapi import requests
 from scrapi import settings
 from scrapi.base import JSONHarvester
 from scrapi.linter.document import RawDocument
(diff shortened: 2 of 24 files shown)


scrapi/harvesters/bhl.py
```diff
@@ -12,7 +12,6 @@ def institution_name_parser(names):
     for inst in names:
         contributor = {
             'name': inst,
-            'email': '',
             'sameAs': []
```
erinspace: get rid of this empty sameAs too?

scrapi/harvesters/crossref.py
```diff
@@ -13,27 +13,26 @@
...
     name = HumanName(author)
-    return {
+    ret = {
         'name': author,
         'givenName': name.first,
         'additionalName': name.middle,
         'familyName': name.last,
-        'email': '',
-        'sameAs': [orcid or '']
+        'sameAs': [orcid] if orcid else []
```
erinspace: else empty list? Or else None?

scrapi/harvesters/bhl.py
erinspace: Oooh or is this taken away by the strip empty function below?

scrapi/processing/cassandra.py
```diff
@@ -32,7 +32,7 @@ def process_normalized(self, raw_doc, normalized):
             contributors=copy_to_unicode(json.dumps(normalized['contributors'])),
             description=copy_to_unicode(normalized.get('description')),
             uris=copy_to_unicode(json.dumps(normalized['uris'])),
-            providerUpdatedDateTime=parse(normalized['providerUpdatedDateTime']),
+            providerUpdatedDateTime=parse(normalized['providerUpdatedDateTime']).replace(tzinfo=None),
```
erinspace: Why strip the tzinfo here, where as in the date_formatter helper you're adding tzinfo? Is this a Cassandra thing?

fabianvf: Yeah, Cassandra thing. cqlengine loses timezone information when it saves datetimes, and since we are moving away from it I just don't think it's worth it.

scrapi/harvesters/bhl.py
fabianvf: No, required fields are not looked into (for now). Empty sameAs is fine for validation and doesn't hurt search though, so I figured I wouldn't take it out (an empty list is different than an empty string- an empty list is basically a null value for lists (it contains no information), while an empty string is actually a value inside string that could have meaning)

scrapi/harvesters/crossref.py
fabianvf: Empty list, indicating that nothing is in the sameAs array. Since this is required, it won't be stripped out by the cleaner.

erinspace:
Cool! Looks good to me! :+1:

fabianvf:
ready for merge then :smile:

PR #334, opened 2015-08-27
No more date errors

fabianvf:
Prevents mapper parsing exceptions in elasticsearch due to unparsed dates.

### DIFF
--- a/scrapi/base/helpers.py
+++ b/scrapi/base/helpers.py
@@ -57,7 +57,7 @@ def compose(*functions):
     credit to sloria
     '''
     def inner(func1, func2):
-        return lambda x: func1(func2(x))
+        return lambda x, *y, **z: func1(func2(x, *y, **z))
     return functools.reduce(inner, functions)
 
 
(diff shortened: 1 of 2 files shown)


scrapi/base/helpers.py
```diff
@@ -198,6 +198,23 @@ def extract_doi_from_text(identifiers):
             continue
 
 
+def null_on_error(task):
+    '''Decorator that makes a function return None on exception'''
+    def inner(*args, **kwargs):
+        try:
+            return task(*args, **kwargs)
+        except Exception:
+            return None
```
erinspace: Can we add a log so we know what's up?

scrapi/base/helpers.py
```diff
@@ -52,12 +52,12 @@ def single_result(l, default=''):
 def compose(*functions):
     '''
     evaluates functions from right to left.
-    ex. compose(f, g)(x) = f(g(x))
+    ex. compose(f, g)(x, *y, **z) = f(g(x, *y, **z))
```
erinspace: As @fabianvf said,  maybe just *x and **y

erinspace:
Sweet!! :+1:

PR #340, opened 2015-08-28
mention autooai utility for writing harvesters

felliott:
@erinspace's autooai tool is mentioned on the SHARE site, and I think it would be helpful to mention it here, too.  I may or may not have spent the better part of yesterday building the NOAA harvester manually because I didn't know about it.  ^_^  D'oh!

Cheers,
Fitz

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ Invoke a harvester a certain number of days back with the ```--days``` argument.
 $ invoke harvester harvester-name --days=5
 ```
 
-Writing a harvester for inclusion with scrAPI?  [autooai](https://github.com/erinspace/autooai) is a utility that makes this easy for most cases.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -202,12 +202,14 @@ For local development, running the ```mit``` harvester is recommended.
...
 Note: harvester-name is the same as the defined harvester "short name".
 
-Invove a harvester a certain number of days back with the ```--days``` argument. For example, to run a harvester 5 days in the past, run:
+Invoke a harvester a certain number of days back with the ```--days``` argument. For example, to run a harvester 5 days in the past, run:
 
 ```bash
 $ invoke harvester harvester-name --days=5
 ```
 
+Writing a harvester for inclusion with scrAPI?  [autooai](https://github.com/erinspace/autooai) is a utility that makes this easy for most cases.
```
erinspace: Hi @felliott! A great addition! One suggestion, can you specify that this is for writing a harvester for sources that have their metadata in OAI-PMH format?

felliott:
Good call, @erinspace.  Updated.

erinspace:
Awesome, thanks!

PR #343, opened 2015-08-31
conditionally create award object

fabianvf:
so that validation errors don't occur due to missing conditionally required components

### DIFF
--- a/scrapi/harvesters/crossref.py
+++ b/scrapi/harvesters/crossref.py
@@ -41,12 +41,12 @@ def process_sponsorships(funder):
     for element in funder:
         sponsorship = {}
 
-        if element['name']:
+        if element.get('name'):
             sponsorship['sponsor'] = {
                 'sponsorName': element['name']
             }
 
-        if element['award']:
+        if element.get('award'):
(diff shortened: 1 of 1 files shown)


scrapi/harvesters/crossref.py
```diff
@@ -37,18 +37,21 @@ def process_contributor(author, orcid):
 
 def process_sponsorships(funder):
     sponsorships = []
+
     for element in funder:
-        sponsorship = {
-            'sponsor': {
+        sponsorship = {}
+
+        if element['name']:
```
erinspace: should this be element.get() to avoid a key error? Same with line 49?

erinspace:
Looks good! This might be a different PR, but maybe as we're changing crossref now it could be added as a quick fix: we're not capturing DOI in the objectUris field, nor is the canonicalUri field repeated in the objectUris.

fabianvf:
True, I'll leave that for a different PR though. We can do an audit of existing harvesters at the beginning of this epic, maybe :)

erinspace:
Ok cool!

PR #344, opened 2015-08-31
add reprs for raw and normalized documents'

fabianvf:
This should make sentry logs way more useful


erinspace:
Nice! Looks good from what I can tell!

PR #347, opened 2015-09-01
Fix errors in repr when serializing for fluentd

fabianvf:
noticed this popping up for some harvesters, now use .get instead of direct access in repr for normalized document

### DIFF
--- a/scrapi/linter/document.py
+++ b/scrapi/linter/document.py
@@ -129,7 +129,7 @@ class NormalizedDocument(BaseDocument):
 
     def __repr__(self):
         return "NormalizedDocument(source='{}', url='{}', providerUpdatedDateTime='{}', ...)".format(
-            self.attributes['shareProperties']['source'],
-            self.attributes['uris']['canonicalUri'],
-            self.attributes['providerUpdatedDateTime']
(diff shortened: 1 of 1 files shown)


erinspace:
Cool! Looks good to me!

scrapi/linter/document.py
```diff
@@ -129,7 +129,7 @@ class NormalizedDocument(BaseDocument):
 
     def __repr__(self):
         return "NormalizedDocument(source='{}', url='{}', providerUpdatedDateTime='{}', ...)".format(
-            self.attributes['shareProperties']['source'],
-            self.attributes['uris']['canonicalUri'],
-            self.attributes['providerUpdatedDateTime']
+            self.attributes.get('shareProperties', {}).get('source'),
+            self.attributes.get('uris', {}).get('canonicalUri'),
+            self.attributes.get('providerUpdatedDateTime')
```
erinspace: although, shouldn't this always exist for Normalized documents? So shouldn't this throw an error if its missing?

fabianvf: yeah, but if it isn't this will throw an error, which requires the ```__repr__``` of the object, which will throw an exception at ```__repr__()``` instead of the original error.

PR #353, opened 2015-09-02
fix typos in clinicaltrials.gov harvester

felliott:
This fixes two small typos in the clinicaltrials.gov harvester.  The typo in the url does not currently seem to cause any problems.  The typo in the xpath expression will cause the eligibility key to always be empty.

Cheers,
Fitz


erinspace:
What! A typo? I never mispell anythign.

PR #354, opened 2015-09-02
[feature] update pubmedcentral OAI-PMH endpoint

felliott:
The original pubmedcentral OAI-PMH url:

  http://www.pubmedcentral.nih.gov/oai/oai.cgi

...now returns a permanent redirect to a new url base:

  http://www.ncbi.nlm.nih.gov/pmc/oai/oai.cgi

This commit updates the harvester and the tests with the new url.  The
tests were updated with a search-and-replace. All pass without further
modification.

### DIFF
--- a/scrapi/harvesters/pubmedcentral.py
+++ b/scrapi/harvesters/pubmedcentral.py
@@ -1,7 +1,7 @@
 """
 Harvester of PubMed Central for the SHARE notification service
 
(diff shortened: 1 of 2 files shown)


fabianvf:
have you run the harvester locally, just checked to make sure the data looks normal?

felliott:
Yep.  Looks correct and passes all tests.

PR #357, opened 2015-09-04
Integrate autooai

fabianvf:
from https://github.com/erinspace/autooai


tasks.py
```diff
@@ -224,3 +224,9 @@ def provider_map(delete=False):
             refresh=True
         )
     print(es.count('share_providers', body={'query': {'match_all': {}}})['count'])
+
+
+@task
+def autooai(shortname, baseurl=None, daterange=None, favicon=False):
+    from scrapi.autooai import generate_oai_harvester
```
erinspace: do we want to add a "missing base url" check and error message?

scrapi/autooai.py
```diff
@@ -0,0 +1,142 @@
...
+        tz_gran = True
+    else:
+        tz_gran = False
+
+    return formatted_oai(ex_call, class_name, shortname, longname, found_url, baseurl, prop_list, tz_gran)
+
+
+def generate_oai_harvester(shortname, baseurl=None, daterange=None, favicon=False):
+
+    if not daterange:
```
erinspace: we can also get rid of dateranges entirely I think, we don't need them anymore for proper generation

scrapi/autooai.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def get_id_props(baseurl):
+    identify_url = baseurl + '?verb=Identify'
+    id_data_request = requests.requests.get(identify_url)
+    id_content = etree.XML(id_data_request.content)
+    return id_content.xpath('//ns0:repositoryName/node()', namespaces=NAMESPACES)[0], id_content.xpath('//ns0:granularity/node()', namespaces=NAMESPACES)[0]
+
+
+def get_favicon(baseurl, shortname):
```
erinspace: I'm torn on continuing to include this or not.... so often, an OAI harvester will just have the standard DSpace favicon for their feeds, leaving you still to have to go search for the original.... your call though!

autooai.py
```diff
@@ -0,0 +1,142 @@
...
+        pre_names = push_request_boundaries(shortname, prop_base, start_date, end_date)
+        all_names = [name.tag.replace('{' + NAMESPACES['dc'] + '}', '') for name in pre_names]
+        return list({name for name in all_names if name not in BASE_SCHEMA}) + ['setSpec']
+
+    # If anything at all goes wrong, just render a blank form...
+    except Exception as e:
+        raise ValueError('OAI Processing Error - {}'.format(e))
+
+
+def push_request_boundaries(shortname, prop_base, startdate, enddate):
```
erinspace: While this function is awesome, we don't really need it if we're gonna get rid of date ranges when querying for elements to add to the properties list, maybe.

autooai.py
```diff
@@ -0,0 +1,142 @@
...
+        tz_gran = True
+    else:
+        tz_gran = False
+
+    return formatted_oai(ex_call, class_name, shortname, longname, found_url, baseurl, prop_list, tz_gran)
+
+
+def generate_oai_harvester(shortname, baseurl=None, daterange=None, favicon=False):
+
+    if not daterange:
```
erinspace: Ah, sorry to repeat maybe, but do we need date ranges any more? We can probably get rid of date altogether!

PR #360, opened 2015-09-10
University of Tennessee Knoxville

fabianvf:


scrapi/harvesters/utnknoxville.py
```diff
@@ -0,0 +1,34 @@
...
+Harvester for the Trace: Tennessee Research and Creative Exchange for the SHARE project
+Example API call: http://trace.tennessee.edu/do/oai/?verb=ListRecords&metadataPrefix=oai_dc
+'''
+from __future__ import unicode_literals
+
+from scrapi.base import OAIHarvester
+
+
+class UtnknoxvilleHarvester(OAIHarvester):
+    short_name = 'utnknoxville'
```
erinspace: Since they call their repository "trace" should we consider that for the shortname? UT is a sorthand we've used for the University of Texas, so I was confused at first.

scrapi/harvesters/utnknoxville.py
```diff
@@ -0,0 +1,34 @@
...
+    long_name = 'Trace: Tennessee Research and Creative Exchange'
+    url = 'http://trace.tennessee.edu'
+
+    base_url = 'http://trace.tennessee.edu/do/oai/'
+    property_list = ['date', 'source', 'identifier', 'type', 'format', 'setSpec']
+    timezone_granularity = True
+
+    approved_sets = [u'utk_accopubs', u'utk_agrieconpubs', u'utk_animpubs', u'utk_anthpubs',
```
erinspace: Thinking about this, this might actually be all the sets.... we should compare against http://trace.tennessee.edu/do/oai/?verb=ListSets, and just not include it if they're the same.

Come to think of it maybe we should build this into autooai... my only thought is if they're different then we should do it, but if they're the same, if they add a new set it'll be filtered by default.

What do you think?

fabianvf: I think they are different- the original VCR I recorded failed tests after I added the approved_sets

erinspace: Oooh excellent way of checking!

erinspace:
Woop!

PR #363, opened 2015-09-11
Make start and end date parsing more sane

erinspace:
closes [#SHARE-31]

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -158,23 +158,8 @@ def harvester(harvester_name, async=False, start=None, end=None):
     if not registry.get(harvester_name):
         raise ValueError('No such harvesters {}'.format(harvester_name))
 
-    if end:
-        if parse(end).date() > date.today():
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -158,8 +158,23 @@ def harvester(harvester_name, async=False, start=None, end=None):
     if not registry.get(harvester_name):
         raise ValueError('No such harvesters {}'.format(harvester_name))
 
-    start = parse(start).date() if start else date.today() - timedelta(settings.DAYS_BACK)
-    end = parse(end).date() if end else date.today()
+    if end:
```
fabianvf: This seems like a lot of logic, would something like:
```python
end = parse(end).date() if end else date.today()
start = parse(start).date() if start else end - timedelta(settings.DAYS_BACK)
```

work?

fabianvf:
:+1:

PR #365, opened 2015-09-14
fix clinical trials serialization errors

fabianvf:


erinspace:
Seems simple enough to me! My only thought is this seems a little unreadable for the sake of leaving it in the schema as one line each. You might consider breaking it up into a helper function like the ```process_contributors``` ones just to make it a little more accessible to read. No need to if you don't want to though!

erinspace:
sweet :+1:

PR #366, opened 2015-09-14
Add 13 harvesters for the Pensoft suite of open journals

erinspace:
Pensoft OAI PMH URLs from this list: http://arphahub.com/about/webservices

One file has 13 seperate harvesters. Some are older than others, and don't have a lot of new content. But it was easier to add them all in one swoop!

Also - made it so the harvester does not need an end date specified - these pensoft OAI endpoints broke when specifying an end date, so making them optional fixes this


scrapi/harvesters/pensoft.py
```diff
@@ -0,0 +1,179 @@
...
+    property_list = ['format', 'rights', 'source', 'relation', 'date', 'identifier', 'type', 'setSpec']
+    timezone_granularity = False
+
+    # don't include the end date for pensoft harvesters
+    def harvest(self, start_date=None, end_date=None):
+
+        start_date = (start_date or date.today() - timedelta(settings.DAYS_BACK)).isoformat()
+        end_date = (end_date or date.today()).isoformat()
+
+        if self.timezone_granularity:
```
fabianvf: Do any of the pensoft harvesters have timezone granularity?

erinspace: Nope... good call

scrapi/harvesters/pensoft.py
```diff
@@ -0,0 +1,179 @@
...
+            end_date += 'T00:00:00Z'
+
+        furled_baseurl = furl(self.base_url)
+        furled_baseurl.args['verb'] = 'ListRecords'
+        furled_baseurl.args['metadataPrefix'] = 'oai_dc'
+        furled_baseurl.args['from'] = start_date
+
+        records = self.get_records(furled_baseurl.url, start_date)
+
+        rawdoc_list = []
```
fabianvf:
```python
return [
    RawDocument({
        'doc': etree.tostring(record, encoding=self.record_encoding),
        'source': self.short_name,  # The util.copy_to_unicode stuff is already done later in the pipeline now
        'docID': record.xpath('ns0:header/ns0:identifier', namespaces=self.namespaces)[0].text,
        'filetype': 'xml'
    }) for record in records
]
```

scrapi/harvesters/pensoft.py
```diff
@@ -0,0 +1,179 @@
...
+        return 'http://{short}.pensoft.net/oai.php/?set={short}'.format(short=self.short_name)
+
+    @property
+    def url(self):
+        return 'http://{short}.pensoft.net'.format(short=self.short_name)
+
+    property_list = ['format', 'rights', 'source', 'relation', 'date', 'identifier', 'type', 'setSpec']
+    timezone_granularity = False
+
+    # don't include the end date for pensoft harvesters
```
fabianvf: Should we get in contact with pensoft about this limitation?

PR #367, opened 2015-09-14
Add prelimniary harvester for OpenAire

erinspace:
There are many ways we can collaborate with OpenAire. While we are figuring those out, here is a simple incorporation of the OpenAire OAI PMH metadata endpoint into SHARE.


erinspace:
Note, this PR will pass tests once #368 is merged (as the test generation needed this no-resume option)

PR #368, opened 2015-09-15
[WIP] Add "resume" argument to harvesters

erinspace:
Add arg that will only grab first page of results. This will help for both local development and test generation, as we can limit the number of results that the test harvesters harvest, and local harvesters take in.

### DIFF
--- a/scrapi/harvesters/dataone.py
+++ b/scrapi/harvesters/dataone.py
@@ -155,12 +155,12 @@ class DataOneHarvester(XMLHarvester):
         'description': ("str[@name='abstract']/node()", single_result)
     }
 
-    def harvest(self, start_date=None, end_date=None):
+    def harvest(self, start_date=None, end_date=None, resume=True):
 
         start_date = start_date or date.today() - timedelta(settings.DAYS_BACK)
         end_date = end_date or date.today()
 
-        records = self.get_records(start_date, end_date)
+        records = self.get_records(start_date, end_date, resume)
 
         xml_list = []
         for record in records:
@@ -175,7 +175,7 @@ class DataOneHarvester(XMLHarvester):
 
         return xml_list
 
-    def get_records(self, start_date, end_date):
+    def get_records(self, start_date, end_date, resume):
         ''' helper function to get a response from the DataONE
         API, with the specified number of rows.
         Returns an etree element with results '''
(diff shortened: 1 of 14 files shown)


fabianvf:
I wonder if it might be better to specify the number of times to follow links? That would allow us to choose the minimum number of jumps for our other harvesters to pass

erinspace:
I like that idea. I'll change it so it does that!

scrapi/harvesters/dataone.py
```diff
@@ -183,4 +183,8 @@ def get_records(self, start_date, end_date):
             docs = etree.XML(data.content).xpath('//doc')
             for doc in docs:
                 yield doc
-            n += 1000
+
+            if page_limit and int(page_limit) == n / 1000:
```
fabianvf: This may have different behavior in python3 (division operator semantics changed slightly, IIRC)

fabianvf: Also, I would put this check as a condition of the while loop.

scrapi/harvesters/harvarddataverse.py
```diff
@@ -87,7 +87,7 @@ def harvest(self, start_date=None, end_date=None):
 
         return record_list
 
-    def get_records(self, search_url):
+    def get_records(self, search_url, page_limit):
```
fabianvf: can you make this into a generator?

scrapi/harvesters/osf.py
```diff
@@ -113,6 +113,10 @@ def get_records(self, search_url):
                 all_records.append(record)
 
             from_arg += 1000
-            records = requests.get(search_url + '&from={}'.format(str(from_arg)), throttle=10)
+
+            if page_limit and int(page_limit) == from_arg / 1000:
```
fabianvf: see above

scrapi/harvesters/plos.py
```diff
@@ -72,9 +72,12 @@ def fetch_rows(self, start_date, end_date):
             for doc in etree.XML(response.content).xpath('//doc'):
                 yield doc
 
-            current_row += self.MAX_ROWS_PER_REQUEST
+            if page_limit and int(page_limit) == current_row / self.MAX_ROWS_PER_REQUEST:
```
fabianvf: Should also be a condition of the outer loop.

scrapi/harvesters/dataone.py
chrisseto:
```python
# python3
1 / 2   # 0.5
1 // 2  # 0

# python 2
1 / 2   # 0
1 // 2  # 0

from __future__ import division
1 / 2   # 0.5
1 // 2  # 0
```

fabianvf: thx

PR #369, opened 2015-09-15
Add support for doctests

fabianvf:
I think the rest of the tests can be added incrementally, but the ability is there now.

### DIFF
--- a/scrapi/registry.py
+++ b/scrapi/registry.py
@@ -3,8 +3,9 @@ import sys
 
 class _Registry(dict):
 
+    # These must be defined so that doctest gathering doesn't make
+    # pytest crash when trying to figure out what/where scrapi.registry is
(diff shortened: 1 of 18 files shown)


scrapi/registry.py
```diff
@@ -3,6 +3,10 @@
 
 class _Registry(dict):
 
+    __file__ = __file__
+
+    __name__ = __name__
+
```
erinspace: Can we put a comment about this?

scrapi/registry.py
```diff
@@ -3,6 +3,10 @@
 
 class _Registry(dict):
 
+    __file__ = __file__
```
erinspace: Can we add a comment about this?

erinspace:
Looks great!! :+1:

PR #370, opened 2015-09-16
hotfix - Fix dataone URI harvesting

erinspace:
- Add use_doi kwarg option to standard URI processor 
- Make the dataone harvester use this kwarg and use the DOIs for the canonicalUri

### DIFF
--- a/scrapi/base/helpers.py
+++ b/scrapi/base/helpers.py
@@ -105,7 +105,7 @@ def format_tags(all_tags, sep=','):
     return list(set([six.text_type(tag.lower().strip()) for tag in tags if tag.strip()]))
 
 
-def oai_process_uris(*args):
+def oai_process_uris(*args, **kwargs):
     identifiers = []
     for arg in args:
         if isinstance(arg, list):
@@ -135,7 +135,10 @@ def oai_process_uris(*args):
                 provider_uris.append(found_url)
 
     try:
-        canonical_uri = (provider_uris + object_uris)[0]
+        if kwargs.get('use_doi') and len(object_uris) > 0:
+            canonical_uri = object_uris[0]
+        else:
+            canonical_uri = (provider_uris + object_uris)[0].strip().replace(']', '')
     except IndexError:
         raise ValueError('No Canonical URI was returned for this record.')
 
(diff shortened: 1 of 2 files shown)


scrapi/base/helpers.py
```diff
@@ -135,7 +135,10 @@ def oai_process_uris(*args):
                 provider_uris.append(found_url)
 
     try:
-        canonical_uri = (provider_uris + object_uris)[0]
+        if kwargs.get('use_doi') and len(object_uris) > 0:
```
fabianvf: Would be better to make this a top-level named argument, rather than put it in kwargs.

erinspace:
Also, misnamed my git branch - it is indeed a hotfix branch though!

scrapi/base/helpers.py
```diff
@@ -135,7 +136,10 @@ def oai_process_uris(*args):
                 provider_uris.append(found_url)
 
     try:
-        canonical_uri = (provider_uris + object_uris)[0]
+        if use_doi and len(object_uris) > 0:
+            canonical_uri = object_uris[0]
```
fabianvf: Are object_uris guaranteed to be DOIs?

fabianvf: Also, do harvesters that use this flag not have any providerUris? The canonicalUri would ideally be some sort of splash page

scrapi/base/helpers.py
```diff
@@ -135,7 +136,10 @@ def oai_process_uris(*args):
...
+        else:
+            canonical_uri = (provider_uris + object_uris)[0].strip().replace(']', '')
```
fabianvf: Is this `.replace` still necessary, or did the regex fix that issue?

scrapi/base/helpers.py
erinspace: A doi oftentimes does point to the sources splash page where the resource lives -- it does on dataone and others I've seen for sure

fabianvf: hmm, I guess there is no way to split out DOIs pointing to splash pages vs elsewhere. Though if DOIs usually point to splash pages, should they be in providerUris instead?

erinspace:
Incorporated all these changes into #372, closing this PR!

PR #372, opened 2015-09-16
Update URI Processing

erinspace:
Closes [#SHARE-54]

A harvester with a long piece of text was incorrectly parsing the URL to include a "]" at the end.

This PR fixes that, and also generally improves URL parsing to also search for dois inside of paragraphs of text.

### DIFF
--- a/scrapi/base/helpers.py
+++ b/scrapi/base/helpers.py
@@ -122,14 +122,21 @@ def gather_identifiers(args):
     return identifiers
 
 
+def maybe_group(match):
+    '''
+    evaluates an regular expression match object, returns the group or none
+    '''
+    return match.group() if match else None
+
+
 def gather_object_uris(identifiers):
     object_uris = []
     for item in identifiers:
         if 'doi' in item.lower():
             url_doi, just_doi = URL_REGEX.search(item), DOI_REGEX.search(item)
-            url_doi = url_doi.group() if url_doi else None
-            just_doi = format_doi_as_url(just_doi.group()) if just_doi else None
-            object_uris.append(url_doi or just_doi)
+            url_doi = maybe_group(url_doi)
+            just_doi = maybe_group(just_doi)
+            object_uris.append(url_doi or format_doi_as_url(just_doi))
 
     return object_uris
 
@@ -139,8 +146,7 @@ def seperate_provider_object_uris(identifiers):
     provider_uris = []
     for item in identifiers:
 
-        found_url = URL_REGEX.search(item)
-        found_url = found_url.group() if found_url else None
+        found_url = maybe_group(URL_REGEX.search(item))
 
         if found_url:
             if 'viewcontent' in found_url:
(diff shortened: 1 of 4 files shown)


scrapi/base/helpers.py
```diff
@@ -118,11 +118,15 @@ def oai_process_uris(*args):
     provider_uris = []
     for item in identifiers:
         if 'doi' in item.lower():
-            doi = item.replace('doi:', '').replace('DOI:', '').strip()
-            if 'http://dx.doi.org/' in doi:
-                object_uris.append(doi)
-            else:
-                object_uris.append('http://dx.doi.org/{}'.format(doi))
+            try:
```
fabianvf: The complexity of this function is getting a little high. Might be a good idea to break some of the functionality into smaller helper functions, just to make the flow of logic a little more transparent.

erinspace: Agreed. Thought it was getting a little out of hand. I'll do that!

scrapi/base/helpers.py
```diff
@@ -114,16 +114,29 @@ def oai_process_uris(*args):
...
-    provider_uris = []
     for item in identifiers:
         if 'doi' in item.lower():
-            doi = item.replace('doi:', '').replace('DOI:', '').strip()
-            if 'http://dx.doi.org/' in doi:
-                object_uris.append(doi)
-            else:
-                object_uris.append('http://dx.doi.org/{}'.format(doi))
-
+            try:
```
fabianvf: Also this nested exception pattern isn't great, is there any way you can make it conditional instead of exception driven?
```python
url_doi, just_doi = URL_REGEX.search(item).group(), DOI_REGEX.search(item).group()
object_uris.append(url_doi or 'http://dx.doi.org/{}'.format(just_doi.replace('doi:', '').replace('DOI:', '').strip())
```
You could also break the `'http://dx.doi.org/{}'.format(just_doi.replace('doi:', '').replace('DOI:', '').strip()` part into a nice helper function for creating DOI urls from dois, could be useful in other applications as well.

scrapi/base/helpers.py
```diff
@@ -114,28 +119,52 @@ def oai_process_uris(*args):
...
-    provider_uris = []
     for item in identifiers:
         if 'doi' in item.lower():
-            doi = item.replace('doi:', '').replace('DOI:', '').strip()
-            if 'http://dx.doi.org/' in doi:
-                object_uris.append(doi)
-            else:
-                object_uris.append('http://dx.doi.org/{}'.format(doi))
+            url_doi, just_doi = URL_REGEX.search(item), DOI_REGEX.search(item)
+            url_doi = url_doi.group() if url_doi else None
```
fabianvf: I think the group function I defined in my previous comment would be a good addition, since this logic is duplicated a few times, and it is a function that could be useful for other harvesters using regex matching.

PR #378, opened 2015-09-18
Update URI processing

erinspace:
Fixes [#SHARE-78] and [#SHARE-79]

### DIFF
--- a/scrapi/base/helpers.py
+++ b/scrapi/base/helpers.py
@@ -172,13 +172,13 @@ def oai_process_uris(*args, **kwargs):
     identifiers = gather_identifiers(args)
     provider_uris, object_uris = seperate_provider_object_uris(identifiers)
 
-    canonical_uri = (provider_uris + object_uris)
+    potential_uri = (provider_uris + object_uris)
     if use_doi:
         for uri in object_uris:
             if 'dx.doi.org' in uri:
-                canonical_uri = [uri]
+                potential_uri = [uri]
     try:
-        canonical_uri = canonical_uri[0]
+        canonical_uri = potential_uri[0]
     except IndexError:
         raise ValueError('No Canonical URI was returned for this record.')
 
--- a/scrapi/base/helpers.py
+++ b/scrapi/base/helpers.py
@@ -137,7 +137,9 @@ def gather_object_uris(identifiers):
             url_doi, just_doi = URL_REGEX.search(item), DOI_REGEX.search(item)
             url_doi = maybe_group(url_doi)
             just_doi = maybe_group(just_doi)
-            object_uris.append(url_doi or format_doi_as_url(just_doi))
+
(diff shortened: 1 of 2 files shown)


scrapi/base/helpers.py
```diff
@@ -131,13 +131,20 @@ def maybe_group(match):
 
 
 def gather_object_uris(identifiers):
+    '''
+    Gathers object URIs if there are any
+    >>> gather_object_uris(['nopenope', 'doi:10.10.gettables', 'http://dx.doi.org/yep'])
+    [u'http://dx.doi.org/10.10.gettables', u'http://dx.doi.org/yep']
```
fabianvf: This will actually fail in python 3 (stupid unicode...)

need to do 
```python
>>> print(gathered[0]
http://dx.doi.org/10.10.gettables
>>> print(gathered[1]
http://dx.doi.org/yep
```
At least until we can drop python 2 support

scrapi/base/helpers.py
```diff
@@ -165,13 +172,13 @@ def oai_process_uris(*args, **kwargs):
     identifiers = gather_identifiers(args)
     provider_uris, object_uris = seperate_provider_object_uris(identifiers)
 
+    canonical_uri = (provider_uris + object_uris)
```
fabianvf: I think candidates or something like that would be a better name for this, the canonicalUri variable should only ever hold the canonicalUri

scrapi/base/helpers.py
```diff
@@ -165,13 +175,13 @@ def oai_process_uris(*args, **kwargs):
     identifiers = gather_identifiers(args)
     provider_uris, object_uris = seperate_provider_object_uris(identifiers)
 
+    potential_uri = (provider_uris + object_uris)
```
fabianvf: SUUPER nitpicky, but since this is a list, can you pluralize it?

scrapi/base/helpers.py
```diff
@@ -183,13 +193,7 @@ def oai_process_uris(*args, **kwargs):
...
 
 def oai_extract_dois(*args):
-    identifiers = []
-    for arg in args:
-        if isinstance(arg, list):
-            for identifier in arg:
-                identifiers.append(identifier)
-        elif arg:
-            identifiers.append(arg)
+    identifiers = gather_identifiers(args)
```
fabianvf: beautiful

scrapi/base/helpers.py
```diff
@@ -202,13 +206,7 @@ def oai_extract_dois(*args):
...
 
 def oai_process_contributors(*args):
-    names = []
-    for arg in args:
-        if isinstance(arg, list):
-            for name in arg:
-                names.append(name)
-        elif arg:
-            names.append(arg)
+    names = gather_identifiers(args)
```
fabianvf: see above

PR #400, opened 2015-10-22
Harvester for Purdue University

erinspace:
Date range harvesting is a little off, so this is in progress for now.


erinspace:
duplicated in #447

PR #401, opened 2015-10-23
NeuroVault harvester

tyarkoni:
Adds a working harvester for NeuroVault collections (neurovault.org). A collection in NeuroVault is a set of images with some common theme; often these are the images that correspond to a single published article. Per some of the other JsonHarvesters, only basic metadata is extracted; we don't store all of the (dozens of) neuroimaging fields included in each collection (though we could). We also don't yet implement specifying date ranges, so at present the harvest() call will grab all available records. Since the number of records is in the 100s, this only takes a few seconds, but we intend to implement date searching in the NeuroVault API shortly.

Please review this and let me know if there are any problems.


scrapi/harvesters/neurovault.py
```diff
@@ -0,0 +1,79 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def process_contributors(authors):
+
+    if authors is None:
+        return []
+    authors = re.split(',\s|\sand\s', authors)
+    return default_name_parser([a for a in authors])
```
fabianvf: You should just be able to pass in authors, I think the list that results from this comprehension is the same as the original.

scrapi/harvesters/neurovault.py
```diff
@@ -0,0 +1,79 @@
...
+            )
+        }
+
+    def harvest(self, start_date=None, end_date=None):
+
+        api_url = self.url + 'api/collections/?format=json'
+
+        record_list = []
+
+        while True:
```
fabianvf: Instead of setting this to `while True`, you could make it `while api_url`, and then you don't need to have the explicit break below either

fabianvf:
Hey, thanks so much for the pull request! It looks like I broke the develop build a little while ago, but I'm working on fixing that. Could you add a favicon for the service at `img/favicons/neurovault_favicon.ico`?

scrapi/harvesters/neurovault.py
```diff
@@ -0,0 +1,79 @@
...
+        return []
+    authors = re.split(',\s|\sand\s', authors)
+    return default_name_parser([a for a in authors])
+
+
+class NeuroVaultHarvester(JSONHarvester):
+    short_name = 'neurovault'
+    long_name = 'NeuroVault.org'
+    url = 'http://www.neurovault.org/'
+    count = 0
```
fabianvf: What is this used for?

scrapi/harvesters/neurovault.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @property
+    def schema(self):
+        return {
+            'contributors': ('/authors', process_contributors),
+            'uris': {
+                'canonicalUri': '/url',
```
fabianvf: Could you duplicate this URL into one of the other URL fields? The other fields provide a bit more information about what the URI points to, the canonicalUri field is just there for convenience.

fabianvf: Could you duplicate this URL into one of the other URL fields? The other fields provide a bit more information about what the URI points to, the canonicalUri field is just there for parsing/ingestion convenience.

scrapi/harvesters/neurovault.py
```diff
@@ -0,0 +1,79 @@
...
+            },
+            'title': '/name',
+            'providerUpdatedDateTime': ('/modify_date', datetime_formatter),
+            'description': '/description',
+            'otherProperties': build_properties(
+                ('owner_name', '/owner_name'),
+                ('doi', '/DOI'),
```
fabianvf: If a DOI exists, it can be formatted to a URI using a resolver, so we prefer to have them in one of the URI fields

fabianvf:
I also noticed a few other URL fields when I was looking through the data, namely `paper_url` and `full_dataset_url`, would it be possible to include these fields as well?

fabianvf:
I was also curious, how persistent are the URLs that are being retrieved?

tyarkoni:
Thanks for all the feedback! I'm swamped this week, but will get back to this early next week.

fabianvf:
No problem! Again, thanks so much for the contribution. Look forward to getting this data in.

scrapi/harvesters/neurovault.py
tyarkoni: Good catch; that was leftover from an earlier attempt that did the splitting in the list comprehension

scrapi/harvesters/neurovault.py
tyarkoni: Nothing, apparently :)

scrapi/harvesters/neurovault.py
tyarkoni: Will do. Is there a doc somewhere that lays out the expected schema? I was trying to infer what the various Uris were from the other Harvesters.

fabianvf: Yep, no problem! [This is the most up to date schema](https://osf.io/wur56/wiki/Schema/?view), and [this is the repository for it](https://github.com/CenterForOpenScience/SHARE-schema). It is in JSON-schema (and the one in the repo is YAML, because it is easier to write by hand). If you have any questions let us know. Also feel free to propose changes in the SHARE-schema repo.

tyarkoni: Nevermind, found it in normalized.json. Maybe add a Development section to README and point to it (and the base Harvester class) there?

scrapi/harvesters/neurovault.py
tyarkoni: Well... it's a bit weird in this case, because the DOI belongs to the article associated with the NeuroVault collection, and not the NeuroVault collection itself. And the article generally doesn't describe or mention the NeuroVault collection (because those are usually uploaded later). Also, currently only a small fraction of collections have DOIs, and if I recall correctly the Uri fields expect a valid URI and throw an exception if None or an empty string is passed. So I'll probably leave it this way for now, unless you have another suggestion.

scrapi/harvesters/neurovault.py
fabianvf: That is a good idea. We do have more documentation at [osf.io/wur56/wiki](https://osf.io/wur56/wiki), but it would be good to mention it in the README.

tyarkoni:
I've added the url fields you mentioned (though as far as I know full_dataset_url has never been used so far). URLs are expected to be permanent, but are under user control. The only thing that would cause a change or removal of a resource is either (a) a database crash and loss of all backups (unlikely), or (b) a user deciding to pull/hide one of their collections (not uncommon). We can't really do anything about the latter, since part of the pitch for NeuroVault is that users have control over their images. But in practice, most collections that are made public (all collections start life out as private, and are inaccessible via API) seem to stick around.

tyarkoni:
Oh, and we don't have a favicon (we don't even have a logo right now). I'll add that once we have something, but not sure when that'll happen. I can add an empty file for now if that helps. @chrisfilo may have other thoughts.

chrisfilo:
Yes we struggled to come up with a sufficient logo. Let's use an empty file for now.

BTW full_dataset_url is being used in some collections. It's suppose to point to other (in most cases RAW) data that correspond to the statistical maps

scrapi/harvesters/neurovault.py
```diff
@@ -0,0 +1,79 @@
...
+    short_name = 'neurovault'
+    long_name = 'NeuroVault.org'
+    url = 'http://www.neurovault.org/'
+
+    @property
+    def schema(self):
+        return {
+            'contributors': ('/authors', process_contributors),
+            'uris': {
+                'objectUris': '/url',
```
fabianvf: This should actually be a list. 
```python
'objectUris': ('/url' : lambda x: [x])
```

fabianvf: Did the validator not catch this? That might be a bug...

scrapi/harvesters/neurovault.py
fabianvf: This is actually a great link to put into `descriptorUris`, I think, which are supposed to point to descriptions of the research object (ie, articles, abstracts, landing pages, etc). I'd recommend something like this:
```python
'uris': {
    ...,
    'descriptorUris': ('/DOI', lambda x: ['http://dx.doi.org/{}'.format(x)]),
}
```

In the search interface we only expose the `canonicalUri`, so the vast majority of people will only ever see that. These other fields are mostly for our own benefit, as we can use the additional identifiers for disambiguation.

tyarkoni: Will add that to the descriptorUris. BTW, I can't add the paper_url and full_dataset_url to the schema. It throws a JsonPointerException; apparently it's detecting them as URLs and then it doesn't like the fact that some (most) records have null values. Not sure if it's the field name or the actual presence of http:// strings that's doing it, but it seems like a pretty strong requirement to force all optional URLs to be set... Will comment them out for now.

tyarkoni: Actually: the issue with the DOIs is that most records don't have a DOI. So if I add the descriptorUris as you propose, we'll end up with a lot of "http://dx.doi.org/None" links, which is probably not desirable. And if I check for DOI before setting, we'll still end up with empty URLs which the validator would reject. Suggestions?

scrapi/harvesters/neurovault.py
```diff
@@ -0,0 +1,80 @@
...
+                'descriptorUris': ('/DOI', lambda x: ['http://dx.doi.org/{}'.format(x)] if x else []),
+                'canonicalUri': '/url',
+            },
+            'title': '/name',
+            'providerUpdatedDateTime': ('/modify_date', datetime_formatter),
+            'description': '/description',
+            'otherProperties': build_properties(
+                ('owner_name', '/owner_name'),
+                ('doi', '/DOI'),
+                # ('paper_url', 'paper_url'),
```
fabianvf: It looks like you are missing the `/` marks on the paths.

scrapi/harvesters/neurovault.py
fabianvf: That seems odd, I'll pull it down and take a look

scrapi/harvesters/neurovault.py
tyarkoni: gah. fixed.

fabianvf:
I added a test response for the harvester here: https://github.com/NeuroVault/scrapi/pull/3

fabianvf:
Cool. Only failure is not your fault

fabianvf:
Actually, think the favicon name needs to be `neurovault_favicon.ico`, instead of `neurovault_favico.ico`. I can fix that post-merge though :+1:

PR #408, opened 2015-11-07
Institutions

hmoco:


institutions/grid.py
```diff
@@ -0,0 +1,45 @@
+import json
+import logging
+from schema_transformer.transformer import JSONTransformer
+
+from .institutions import Institution
+
+logger = logging.getLogger(__name__)
+
+class GridTransformer(JSONTransformer):
```
fabianvf: You can just use the JSONTransformer, no need to subclass it

institutions/ipeds.py
```diff
@@ -0,0 +1,53 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class IpedsTransformer(CSVTransformer):
+    def _transform_string(self, val, doc):
+        val = super(IpedsTransformer, self)._transform_string(val, doc)
+        return six.u(val)
+
+    def load(self, doc):
```
hmoco: A ` CSVTransformer ` raises an ` ImplementationError ` for a not defined ` load ` function.

fabianvf: What moron wrote that?
lol jk, sorry about that.

api/webview/urls.py
```diff
@@ -7,5 +7,6 @@
     url(r'^documents/status', views.status, name='status'),
     url(r'^documents/(?P<source>\w+)/$', views.DocumentsFromSource.as_view(), name='source'),
     url(r'^documents/(?P<source>[a-z]+)/(?P<docID>(.*))/$', views.document_detail, name='document_detail'),
+    url(r'^institutions/', views.institution_detail),
```
erinspace: The name is used to reference it elsewhere, like from a template with the url tag - {% url name_goes_here %} so you don't neeeeed one if you're not going to reference it from a django template. But if you are, maybe, might as well give it a name

fabianvf:
Oh, can you add an institutions section to the README, which explains how to get everything set up (including finding the datasets and stuff?)

tasks.py
```diff
@@ -302,3 +303,32 @@ def reset_all():
...
+    if not skip:
+        resp = six.moves.input('You are about to delete the institutions index. Are you sure? (y, n)\n')
+        if resp not in ('y', 'Y', 'Yes', 'yes'):
+            print('Remove institutions stopped.')
+            return
+    from institutions.institutions import remove
+    remove()
+
+
+def grid(grid_file):
```
fabianvf: Doesn't this make it so you can't `inv grid` with out a file? also, I'd call it `filepath`, so you don't have to remember whether it's `-i` or `-g`

README.md
```diff
@@ -334,3 +334,15 @@ To fix:
 - reinstall requirements as usual
 
 Answer found in [this stack overflow question and answer](http://stackoverflow.com/questions/23172384/lxml-runtime-error-reason-incompatible-library-version-etree-so-requires-vers)
+
+### Institutions!
+Scrapi supports the addition of institutions in a separate index (` institutions `). Unlike data stored in the ` share ` indices, institution's metadata is updated
+much less frequently, meaning that simple parsers can be used to manually load data from providers instead of using scheduled harversters.
```
fabianvf: `harversters` -> `harvesters`

institutions/grid.py
```diff
@@ -0,0 +1,52 @@
...
+    'web_url': ('/links', lambda x: x[0] if x else None),
+    'id_': '/id',
+    'other_names': ('/aliases', '/acronyms', lambda x, y: x + y if x and y else x if x else y if y else None)
+}
+
+
+def get_jsons(grid_file):
+    with open(grid_file) as f:
+        f.readline()  # Pop off the top
+        f.readline()
```
fabianvf: Are there supposed to be two readlines here?

institutions/grid.py
```diff
@@ -0,0 +1,52 @@
...
+    transformer = JSONTransformer(schema)
+    for doc in get_jsons(grid_file):
+        transformed = transformer.transform(doc, load=False)
+        try:
+            # Prevent logger output encoding errors from stopping script
+            logger.info('Adding {0}.'.format(transformed['name']))
+        except Exception:
+            pass
+        for key, val in six.iteritems(transformed):
+            if type(val) is bytes:
```
fabianvf: I think `isinstance(val, six.binary_type)` would work as well. Does this work in python 2 and 3?

tasks.py
```diff
@@ -302,3 +303,32 @@ def reset_all():
...
+
+
+@task
+def institutions(grid_file='institutions/grid_2015_10_09.json', ipeds_file='institutions/hd2013.csv'):
+    grid(grid_file)
+    ipeds(ipeds_file)
+
+
+@task
+def remove_institutions(skip=False):
```
fabianvf: `force` would make more sense than `skip`.

tasks.py
hmoco: grid and ipeds are helper functions for institutions, and as such are not tasks and not invokable.

institutions/grid.py
hmoco: I believe so? Can't recall why though ...

institutions/grid.py
hmoco: Neither compiler throws an error. However, it is only triggered in python 3 I believe, where byte vals were throwing errors. Will try  ` isinstance `.

PR #412, opened 2015-11-10
[Ready for Review] Push API harvester

fabianvf:
Skips testing for now (it was pretty hard to do)


scrapi/harvesters/push_api.py
```diff
@@ -0,0 +1,108 @@
...
+            response = requests.get(response['next']).json()
+        else:
+            break
+
+
+def gen_harvester(shortname=None, longname=None, url=None, favicon_dataurl=None, **kwargs):
+    assert shortname and longname and url and favicon_dataurl
+    logger.info('Generating harvester {}'.format(shortname))
+
+    from scrapi.processing.elasticsearch import DatabaseManager
```
erinspace: Does this make having elasticsearch setup mandatory? What happens if we remove elasticsearch in the future?

PR #413, opened 2015-11-10
Update utaustin url and approved sets

erinspace:


fabianvf:
Looks like you need to recreate the VCR, the new approved sets don't allow any of the documents in the current VCR to normalize.

PR #415, opened 2015-11-12
Updates to README.md

felliott:
I was reading through the docs and realized we didn't have any instructions for installing and setting up Postgres.  I've added some quick and simple directions, and fixed some small typos / formatting errors on the way.  I've also updated the `brew install elasticsearch` invocation to get v1.7 from the versions keg.  There's no JIRA ticket for this, this is a drive-by PR!


erinspace:
Thanks @felliott !!!

PR #428, opened 2015-11-24
Update to latest grids and ipeds files

hmoco:
Update docs and path to point to the new files, and giving better directions to getting ipeds information.


fabianvf:
Can you also provide documentation for how to deal with the zip file at the command line?

PR #437, opened 2015-12-14
Feature/add harvester

kms6bn:
added Érudit scraper

### DIFF
--- a/scrapi/harvesters/erudit.py
+++ b/scrapi/harvesters/erudit.py
@@ -1,7 +1,8 @@
+# coding=utf-8
 '''
 Harvester for the Erudit for the SHARE project
 
-Example API call: http://oai.erudit.org/oai/?verb=ListRecords&metadataPrefix=oai_dc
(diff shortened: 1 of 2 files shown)


scrapi/harvesters/erudit.py
```diff
@@ -1,8 +1,7 @@
-# coding=utf-8
```
fabianvf: What are these changes for?

fabianvf:
Looks good to me, but it looks like you committed some changes to the Erudit harvester, and you will also need to include a favicon and test response. Also, I just realized I may not have linked you to our documentation, which is here: https://osf.io/wur56/wiki/home/

In particular, the creating a harvester documentation: https://osf.io/wur56/wiki/Creating%20a%20Harvester/

scrapi/harvesters/erudit.py
kms6bn: probably not the best way to correct for this, but I just copied the previous version to overwrite my changes and added it to the pull request. Will try to revert next time.

Also changed erudit to done to avoid confusion later.

Sorry!

PR #444, opened 2015-12-18
Hotfix/fix addis ababa

kms6bn:
updated the uris to read correctly.

however, the identifier is still not a correct link - do I need to update that as well?
image


scrapi/harvesters/addis_ababa.py
```diff
@@ -6,13 +6,38 @@
...
 
+def format_uris_addis_ababa(*args):
+    identifiers = helpers.gather_identifiers(args)
+    provider_uris, object_uris = helpers.seperate_provider_object_uris(identifiers)
+
+    for arg in args:
+        if arg and 'http://hdl.handle.net/123456789/' in arg[0]:
+            DOC_ID = arg[0].replace('http://hdl.handle.net/123456789/', '')
+            canonical_uri = 'http://etd.aau.edu.et/handle/123456789/' + DOC_ID
+            provider_uris = 'http://etd.aau.edu.et/handle/123456789/' + DOC_ID
```
erinspace: Looks like this will make provider_uris into a single URL - should be a list

scrapi/harvesters/addis_ababa.py
```diff
@@ -6,13 +6,38 @@
...
 from scrapi.base import OAIHarvester
+from scrapi.base import helpers
 
+def format_uris_addis_ababa(*args):
+    identifiers = helpers.gather_identifiers(args)
+    provider_uris, object_uris = helpers.seperate_provider_object_uris(identifiers)
+
+    for arg in args:
+        if arg and 'http://hdl.handle.net/123456789/' in arg[0]:
+            DOC_ID = arg[0].replace('http://hdl.handle.net/123456789/', '')
```
erinspace: usually in python variables in ALL_CAPS are constants, so this would be better as lower case (super small detail)

erinspace:
Oh also  -- looks like you made this PR against develop -- you'll need to close it and reopen a new one against master for a hotfix (I do this allll the time)

PR #446, opened 2015-12-18
Hotfix/fix addis ababa

kms6bn:


scrapi/harvesters/addis_ababa.py
```diff
@@ -6,6 +6,32 @@
...
+
+
+def oai_process_uris_addis_ababa(*args):
+
+    identifiers = helpers.gather_identifiers(args)
+    provider_uris, object_uris = helpers.seperate_provider_object_uris(identifiers)
+
+    potential_uris = (provider_uris + object_uris)
+
+    for uri in potential_uris:
```
fabianvf: You can actually get the index and the item at the same time using `enumerate`:
```python
for i, uri in enumerate(potential_uris):
    # i is the index, uri is your uri string
    # do stuff
```

kms6bn:
I think this is what @erinspace had in mind - @fabianvf let me know if there is a simpler way to do this!

scrapi/harvesters/addis_ababa.py
```diff
@@ -6,6 +6,35 @@
...
+def oai_process_uris_addis_ababa(*args):
+
+    identifiers = helpers.gather_identifiers(args)
+    provider_uris, object_uris = helpers.seperate_provider_object_uris(identifiers)
+
+    potential_uris = (provider_uris + object_uris)
+
+    for i, uri in enumerate(potential_uris):
+        if 'http://hdl.handle.net/123456789/' in uri:
+            doc_id = potential_uris[i].replace('http://hdl.handle.net/123456789/', '')
```
fabianvf: You could also just do;
```python
potential_uris[i].replace('http://handle.net/123456789/', 'http://etd.aa.edu.et/handle/123456789')
```
and drop the string concatenation, since all you are really doing with that is replacing the beginning of the string.

kms6bn:
makes sense - updated.

scrapi/harvesters/addis_ababa.py
```diff
@@ -6,6 +6,33 @@
...
+        potential_uris[i].replace('http://handle.net/123456789/', 'http://etd.aa.edu.et/handle/123456789')
+
+    try:
+        canonical_uri = potential_uris[0]
+    except IndexError:
+        raise ValueError('No Canonical URI was returned for this record.')
+
+    provider_uris = potential_uris[0:len(provider_uris)]
+
+    object_uris = potential_uris[(len(provider_uris) + 1):len(potential_uris)]
```
fabianvf: Shouldn't this one just get the rest? It shouldn't need the second part of the slice, unless I am mistaken.

scrapi/harvesters/addis_ababa.py
```diff
@@ -6,6 +6,33 @@
...
+
+    for i, uri in enumerate(potential_uris):
+        potential_uris[i].replace('http://handle.net/123456789/', 'http://etd.aa.edu.et/handle/123456789')
+
+    try:
+        canonical_uri = potential_uris[0]
+    except IndexError:
+        raise ValueError('No Canonical URI was returned for this record.')
+
+    provider_uris = potential_uris[0:len(provider_uris)]
```
fabianvf: Does this have an off-by-one error? You might want `len(provider_uris) -1`.
Alternatively, you could alter the content of `provider_uris` and `object_uris` before concatenating them together, which would mean that you don't have to re-form these two lists at all.

kms6bn:
let me know if this is what you had in mind.

fabianvf:
yep :+1: 
Will merge when the build passes

kms6bn:
It looks like the links are still not working from osf.io/share - any idea why?

image

erinspace:
@kms6bn I bet @fabianvf hasn't done a release yet -- should update when he does so

kms6bn:
@erinspace makes sense, thanks!

kms6bn:
@erinspace @fabianvf it still looks like the links are not working - has this been released yet?

PR #450, opened 2015-12-28
add rcaap harvester, test, and favicon

kms6bn:
Add RCAAP (sorry, I misspelled it in the branch name)


fabianvf:
No problem, but for future reference you can do `git branch -m <new-branch-name>` to rename a branch.

scrapi/harvesters/rcaap.py
```diff
@@ -0,0 +1,19 @@
...
+Example API call: http://www.rcaap.pt/oai?verb=ListRecords&metadataPrefix=oai_dc
+'''
+from __future__ import unicode_literals
+
+from scrapi.base import OAIHarvester
+
+
+class RcaapHarvester(OAIHarvester):
+    short_name = 'rcaap'
+    long_name = 'RCAAP - Repositório Científico de Acesso Aberto de Portugal'
```
fabianvf: I think this file is interpreted as `latin-1` by default in python2, so you'll need to change this character or declare the encoding of the file to be `utf-8`

https://www.python.org/dev/peps/pep-0263/

kms6bn:
Let me know if this addresses the issue. Also - it looks like it now doesn't pass the Travis build, any idea why?

PR #452, opened 2015-12-29
fix uris

kms6bn:


kms6bn:
it looks like vatech fixed their links - I am closing this pull request

PR #453, opened 2015-12-29
updated favicons

kms6bn:
Updated favicons that were D for DSpace and a few others that were incorrect.


kms6bn:
@fabianvf is this good to pull in?

PR #455, opened 2016-01-07
Feature/add elife

kms6bn:
@fabianvf wanted to have you take a look at this before you head out tomorrow - any feedback would be appreciated! 

Also - right now, I have that it just goes to the first page of commits, but if you searched over all time - there are a total of 3 pages and the scraper would only read the first one. Not sure how to deal with that. I also had difficulty getting the license into the correct format.


scrapi/harvesters/elife.py
```diff
@@ -0,0 +1,171 @@
...
+def title_parser(title):
+    parsed_title = ''.join(title)
+    return parsed_title
+
+
+def description_parser(description):
+    try:
+        if "DOI:" in description[-3]:
+            del description[-3:]
+    except:
```
fabianvf: Don't use bare except blocks, they can be pretty dangerous.

scrapi/harvesters/elife.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def title_parser(title):
+    parsed_title = ''.join(title)
+    return parsed_title
+
+
+def description_parser(description):
+    try:
+        if "DOI:" in description[-3]:
```
fabianvf: Might want to reconsider naming here, `description` sounds like a string to me, but it seems like a list. Also, could you explain what this function does?

scrapi/harvesters/elife.py
```diff
@@ -0,0 +1,171 @@
...
+            'name': ('//publisher-name/node()', single_result)
+        },
+        'subjects': '//article-meta/article-categories/descendant::text()',
+        'freeToRead': {
+            'startDate': ('//article-meta/pub-date[@publication-format="electronic"]/*/node()',
+                          elife_date_parser)
+        },
+        'tags': '//kwd/text()',
+        'otherProperties': build_properties(
+                ('license', ('//permissions/license/license-p/ext-link/text()', single_result))
```
fabianvf: We actually have a license field in the SHARE schema.

scrapi/harvesters/elife.py
kms6bn: description is a list taken from the abstract that xpath parses. The issue is that occasionally there is an <italic> or <bold> tag, and it would stop reading once it hit a tag. However, when I did a workaround, it read too much, and to account for that, I removed the extra information. I can play around a little more with xpath to see if there is an easier way.

scrapi/harvesters/elife.py
```diff
@@ -0,0 +1,171 @@
...
+                'filetype': 'xml',
+                'source': self.short_name,
+                'doc': etree.tostring(record),
+                'docID': record.xpath('//article-id[@*]')[0].text,
+            }) for record in xml_records
+        ]
+
+    schema = {
+        'uris': {
+            'canonicalUri': ('//article-id/node()', compose('http://dx.doi.org/10.7554/eLife.{}'.format,
```
fabianvf: Could you also duplicate this in one of the other uri fields, so we can know the type.

scrapi/harvesters/elife.py
kms6bn: I was having difficulty getting it in the schema format without it being rejected - do you know of an example with a custom license?

scrapi/harvesters/elife.py
kms6bn: Can you clarify?

scrapi/harvesters/elife.py
```diff
@@ -0,0 +1,171 @@
...
+    BASE_DATA_URL = 'https://raw.githubusercontent.com/elifesciences/elife-articles/master/{}'
+
+    def harvest(self, start_date=None, end_date=None):
+        start_date = start_date or datetime.date.today() - datetime.timedelta(settings.DAYS_BACK)
+        end_date = end_date or datetime.date.today()
+
+        shas = fetch_commits(self.BASE_URL, start_date.isoformat(), end_date.isoformat())
+
+        files = []
+        for sha in shas:
```
fabianvf: I think something like this might be a little easier to read
```python
files = list({
    fetch_file_names(self.BASE_COMMIT_URL, sha) 
    for sha in shas
})
```

scrapi/harvesters/elife.py
```diff
@@ -0,0 +1,171 @@
...
+            files = files + fetch_file_names(self.BASE_COMMIT_URL, sha)
+            files = list(set(files))
+
+        xml_records = []
```
fabianvf: You can use a list comprehension here as well - 
```python
xml_records = [
    fetch_xml(self.BASE_DATA_URL, filename) 
    for filename in files
]
```

fabianvf:
First round done :+1:
Check out the failing test on travis as well, it looks like it's just a string handling error.

scrapi/harvesters/elife.py
kms6bn: I figured out how to fix this here

scrapi/harvesters/elife.py
kms6bn: I figured out how to fix this here

kms6bn:
The travis error seems to do with encoding between python 2.7 and 3 - have you encountered this before? I have been scripting with python 3.

scrapi/harvesters/elife.py
kms6bn: cool - this worked

scrapi/harvesters/elife.py
kms6bn: Still having trouble getting it to accept the license as part of the schema - I listed it as rights in other properties for now (which is the case for most of the harvesters currently)

scrapi/harvesters/elife.py
kms6bn: this doesnt produce what the code needs (this creates a list of lists - but I needed a single list of all of the file names)

fabianvf: Ah,  try:
```python
from itertools import chain

files = list(set(chain.from_iterable([
    fetch_file_names(self.BASE_COMMIT_URL, sha) 
    for sha in shas
))
```

kms6bn: cool - that worked

scrapi/harvesters/elife.py
```diff
@@ -0,0 +1,175 @@
...
+
+    return parsed_contributors
+
+
+def elife_date_parser(date):
+    date_form = datetime.datetime(int(date[2]), int(date[1]), int(date[0]), 0, 0)
+    return date_form.date().isoformat()
+
+
+# will need to update this to pull from multiple pages
```
kms6bn: @fabianvf should I worry about this? Basically, if you tried to look at two years worth of commits - it would only be reading the first page.

fabianvf: If it's not hard, I'd prefer to have it working. We may want to go back that far at some point.

kms6bn:
This script is now updated to read as many pages as there is results. Currently to get this to run multiple pages I have to input my username and personal access token (you can only access 60 requests per hour without using this, which is not sufficient for larger queries of time, like a 6 month window for example).

```
    $ curl -u username:token https://api.github.com/repos/elifesciences/elife-articles/commits
```

I am not sure how this would work in the SHARE framework?

kms6bn:
@erinspace just wanted to mention that you will need to incorporate the script above and the github API key that I emailed you a week or so back, let me know if you have issues!

erinspace:
@kms6bn yep absolutely, thank you for the reminder! EXCELLENT work on this one, super exciting that it's ready to go live!

PR #456, opened 2016-01-11
Feature/add ut chattanooga

kms6bn:
Adding harvester, favicon, and test for ut chattanooga


scrapi/harvesters/ut_chattanooga.py
```diff
@@ -0,0 +1,29 @@
...
+
+    approved_sets = [
+        u'honors-theses',
+        u'theses',
+        u'student-research',
+        u'ebm-symposium',
+        u'conferences',
+        u'journals',
+        u'jafh'
+    ]
```
kms6bn: @erinspace they had sent us the following:
Approved sets:
[u'honors-theses', u'theses']

based on your recent comment, should I remove the other approved sets I through in there? I had added them since I thought that was how to handle the error that the harvester pops out.

erinspace: Hi @kms6bn ! Yep, I'd go ahead and remove the ones you added.  We usually let the providers let us know what they'd like in SHARE as they know best what they have (as some would prefer undergrad or other content not make it).

kms6bn: @erinspace when I removed those items, it failed travis - have you encountered this?

erinspace:
Yep -- should fail locally now too if you run it.  Looks like you made the yml file with the additional sets, and so it's looking for those in the results, but because they're no longer permitted, it's failing. You'll need to delete the old and then remake the yml file, maybe using different dates to gather a time when there were results from the approved sets. 

This is most definitely annoying and inconvenient for generating test files, as it includes some guess and check to find dates where we get the approved sets. We'll definitely have to refactor some things in the future. I can also give you more details when I'm back in the office tomorrow!

kms6bn:
@erinspace did you ever ask them if they only wanted to include theses and honor theses? Seems very bare-boned

erinspace:
@kms6bn just emailed them, will let you know when we hear back!

PR #460, opened 2016-01-14
Update README: SPRINGER_KEY changed to SPRINGER_API_KEY

kms6bn:
SPRINGER_KEY should be SPRINGER_API_KEY to match variable name in springer.py


erinspace:
Sweet, thanks!

PR #461, opened 2016-01-15
hotfix/replace favicon

kms6bn:
The current favicon for trinity is not representative of Trinity University. I have replaced it with their correct favicon.


erinspace:
Looks good, thanks!

PR #432, opened 2015-12-03
Custom harvester for United States Geological Survey

ZobairAlijan:

PR #471, opened 2016-01-28
[USGS]

MerlinZhang:
Modified USGS harvester from PR #432 (PEP8 fix)


erinspace:
@MerlinZhang Nice! Looks like you have ```scrapi/processing/cassandra.py``` still modified tho... I'd look into putting those back the way they were.

Also if you could add to your PR description that you're also modifying ```travis-dist``` and ```local-dist``` just so we have record of what's changing

PR #476, opened 2016-02-10
Feature/usgs

MerlinZhang:
Fixed USGS harvester (pep8 and unused imports)
Included favicon
Included yaml file


kms6bn:
@MerlinZhang it looks like the description needs to be parsed (still includes html tags)

erinspace:
We usually keep the raw data as it is, and then leave whatever sources use the data to fix the parsing. For example, osf.io/share sanitizes html before displaying!

PR #477, opened 2016-02-11
[dataONE]

MerlinZhang:
Fixed dataONE to stop harvesting results with no information
Included conditional statement to collect only metadata, no data or resources to fix the problem


erinspace:
A note - we'll have to renormalize everything from dataone to put this into full effect which will cut down quite a bit on the total number of documents in SHARE. 

We'll deploy this now so that it applies to all future harvests.

PR #480, opened 2016-02-12
added note about one_test

kms6bn:
just wanted to add a note about the new invoke one_test method


kms6bn:
also added a note on the SHARE documentation here [https://osf.io/wur56/wiki/Creating%20a%20Harvester/]

PR #482, opened 2016-02-16
Feature/biomedcentral

MerlinZhang:


erinspace:
Just a note to explain what this is doing -- it now looks like biomedcentral is using Springer's API to expose its content. So, this PR replaces the old biomedcentral code with similar code to springer's harvester, and only normalizes those with the publisher "biomedcentral."

scrapi/harvesters/biomedcentral.py
```diff
@@ -6,134 +6,128 @@
...
 
 
-def process_contributors(authors):
+def process_urls(urls):
+    all_uris = [url['value'] for url in urls]
 
-    authors = authors.strip().replace('<span class="author-names">', '').replace('</span>', '')
-    authors = authors.split(',')
+    processed_uris = {
+        'canonicalUri': all_uris[0],
```
erinspace: Can you fix this so that the DOI given shows up as the canonicalUri? If you check out the results, the canonical URL should lead to the biomedcentral splash page.

scrapi/harvesters/biomedcentral.py
```diff
@@ -6,134 +6,128 @@
...
-                ('articleFullUrl', '/articleFullUrl'),
-                ('isFree', '/isFree'),
-                ('isHighlyAccessed', '/isHighlyAccessed'),
-                ('status', '/status'),
-                ('abstractPath', '/abstractPath'),
-                ('journal Id', '/journal Id'),
-                ('article_host', '/article_host'),
-                ('longCitation', '/longCitation'),
-                ('is_subscription', '/is_subscription')
+                ('url', '/url'),
```
erinspace: is this URL captured in one of our URL fields? If so we can remove it from otherProperties

erinspace:
Ok! @MerlinZhang pass 1 finished!

PR #483, opened 2016-02-16
[springer]

MerlinZhang:
Updated springer harvester to ignore biomed central
Updated springer to log when a source is not biomed central or springer


erinspace:
Looks good! :+1:

PR #485, opened 2016-02-17
Use json serializers for raw and normalized

erinspace:
Use django's json serializers to display raw and normalized as actual json


erinspace:
This will need an update on the OSF side for how we de-serialize this new format for display on osf.io/share for the raw and normalized data.

PR #503, opened 2016-03-10
Feature/queryurlbyproviderupdatetime

jgw4sq:
I created a new view so that the user can display specific documents restrained by providerupdatetime via url query.

I wrote a new test case in order to confirm that it works as expected.

### DIFF
--- a/api/webview/urls.py
+++ b/api/webview/urls.py
@@ -7,7 +7,7 @@ urlpatterns = [
     url(r'^documents/status', views.status, name='status'),
     url(r'^documents/(?P<source>\w+)/$', views.DocumentsFromSource.as_view(), name='source'),
     url(r'^documents/(?P<source>[a-z]+)/(?P<docID>(.*))/$', views.document_detail, name='document_detail'),
-    url(r'^documents/(?P<from>\d{4}-\d{2}-\d{2})&(?P<until>\d{4}-\d{2}-\d{2})/$', views.DocumentsByProviderUpdatedDateTime.as_view(), name='providerupdate'),
+    url(r'^documents/from=(?P<from>\d{4}-\d{2}-\d{2})&until=(?P<until>\d{4}-\d{2}-\d{2})/$', views.DocumentsByProviderUpdatedDateTime.as_view(), name='providerupdate'),
     url(r'^institutions', views.institutions, name='institutions'),
     url(r'^robots\.txt$', include('robots.urls')),
 ]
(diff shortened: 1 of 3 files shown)


api/webview/urls.py
```diff
@@ -7,6 +7,7 @@
     url(r'^documents/status', views.status, name='status'),
     url(r'^documents/(?P<source>\w+)/$', views.DocumentsFromSource.as_view(), name='source'),
     url(r'^documents/(?P<source>[a-z]+)/(?P<docID>(.*))/$', views.document_detail, name='document_detail'),
+    url(r'^documents/from=(?P<from>\d{4}-\d{2}-\d{2})&until=(?P<until>\d{4}-\d{2}-\d{2})/$', views.DocumentsByProviderUpdatedDateTime.as_view(), name='providerupdate'),
```
erinspace: Instead of adding a new URL route, can you add this optional query parameter to the documents/ route?

api/webview/views.py
```diff
@@ -44,7 +45,25 @@ def perform_create(self, serializer):
...
+    permission_classes = (permissions.IsAuthenticatedOrReadOnly,)
+
+    def perform_create(self, serializer):
+        serializer.save(source=self.request.user)
+
+    def get_queryset(self):
+        """ Return queryset based on provider update time
+        """
+        queryset = Document.objects.all()
+        queryset = queryset.filter(providerUpdatedDateTime__gte=parse(self.kwargs['from'])).filter(providerUpdatedDateTime__lte=parse(self.kwargs['until']))
```
erinspace: This is awesome! Can you modify it so that instead of a new view, it checks for the argument in the normal document list view? It should add a filter only if the query parameter is there, and do nothing if it isn't.  Check out the source one that gloria added in her PR. 

Also, you can go ahead and add a filter on the initial queryset, instead of first querying for everything and then adding a filter (it will be faster!)

so, something like:

```
filters = {}
if self.kwargs.get('from', None):
    filters['from'] = self.kwargs['from']
   # other optional filters here

return queryset(**filters)
````

File: tests/test_api_views.py [hunk 3]
```diff
@@ -60,4 +61,42 @@ def test_institutions(self):
...
+            '/documents/'
+        )
+        response = view(request)        
+        self.assertNotContains(response, "bad",
+                            status_code=200)
+        self.assertContains(response, "good", status_code=200)
+        print(response)
+
+
+class APIViewTests2(APITestCase):
```
erinspace: Once you incorporate your code into the document List view, you shouldbe able to add these tests back into the first class - no need for a second

tests/test_api_views.py
```diff
@@ -60,4 +61,42 @@ def test_institutions(self):
...
+        self.assertContains(response, "rightontime", status_code=200)
+        self.assertNotContains(response, "toolate", status_code =200)
+
+
+def create_new_document(source,providerUpdatedDateTime):
+        return Document.objects.create(source = source, providerUpdatedDateTime = providerUpdatedDateTime)
+def create_document(source,normalized):
+        return Document.objects.create(source= source,normalized= normalized)
+
+
```
erinspace: You can remove these newlines at the bottom and just leave one!

erinspace:
@jgw4sq code review pass 1 complete!

PR #507, opened 2016-03-11
New commit for fixing formating of nist harvester tags

jgw4sq:
This is the new commit for the nist harvester fix, with only the proper files in the commit.

### DIFF
--- a/scrapi/harvesters/nist.py
+++ b/scrapi/harvesters/nist.py
@@ -6,13 +6,39 @@ Example API call: https://materialsdata.nist.gov/dspace/oai/request?verb=ListRec
 from __future__ import unicode_literals
 
 from scrapi.base import OAIHarvester
+from scrapi.base.helpers import updated_schema
 
 
 class NistHarvester(OAIHarvester):
     short_name = 'nist'
     long_name = 'NIST MaterialsData'
     url = 'https://materialsdata.nist.gov'
-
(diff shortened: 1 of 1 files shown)


scrapi/harvesters/nist.py
```diff
@@ -6,13 +6,39 @@
...
+            tags.extend(tag.split("::"))
+        if "," in tag:
+            tags.extend(tag.split(","))
+        elif "::" not in tag:
+            tags.append(tag)
+
+    for tag in tags:
+        if "::" in tag:
+            tags.remove(tag)
+        if tag == "":
```
jeffreyliu3230: `if not tag:` might be better.

scrapi/harvesters/nist.py
```diff
@@ -6,13 +6,39 @@
...
+        if "::" in tag:
+            tags.extend(tag.split("::"))
+        if "," in tag:
+            tags.extend(tag.split(","))
+        elif "::" not in tag:
+            tags.append(tag)
+
+    for tag in tags:
+        if "::" in tag:
+            tags.remove(tag)
```
jeffreyliu3230: Can you change it so that "::" and "" are not added to tags in the first place?

jeffreyliu3230:
Otherwise looks good. Just that it might be more efficient to not add "::" and "" to tags in the first place.

PR #509, opened 2016-03-11
Feature/lastharvest

MerlinZhang:
New Feature

Keeps track of when sources were most recently harvested.
A new database model called LastHarvest keeps records of each source and the most recent date when it was harvested. The database is updated every time a new document from a more recent date is normalized.
Code updates include a new view and url to view the database information for most recently updated sources.


scripts/last_harvest.py
```diff
@@ -0,0 +1,29 @@
+# from api.webview.models import Document
```
jeffreyliu3230: This file needs to be removed from the git repository.

jeffreyliu3230:
Can you add tests to this feature? We normally would require tests to be implemented for every new feature.

PR #513, opened 2016-03-24
update readme for under ubuntu

wearpants:


erinspace:
Thank you so much @wearpants!

PR #516, opened 2016-04-01
Hotfix/umontreal

kms6bn:
removed date from properties list. Also updated language to be read in correctly according to schema.


scrapi/harvesters/umontreal.py
```diff
@@ -7,14 +7,31 @@
...
+from scrapi.base.helpers import updated_schema
 from scrapi.base import OAIHarvester
 
 
+def umontreal_language_processor(languages):
+
+    if not languages:
+        languages = []
+
+    return languages
```
jeffreyliu3230: In the original schema, each item in the languages array is transformed to a ISO 639-3 code, for example, 'English' becomes 'eng' and 'Russian' becomes 'rus'. I checked the source and it seems that they were not correctly coded (for example it uses 'en' instead of 'eng'). There is a function get_code in helpers.py that can do that for you. Can you implement it in your function to return the correct codes?

kms6bn: Below is the description in the schema:
- "description": "The primary languages in which the content of the resource is presented. Values used for this element MUST conform to ISO 639\u20133. This offers two and three letter tags e.g. \"en\" or \"eng\" for English and \"en-GB\" for English used in the UK."

You can find all the tags here: https://www.loc.gov/standards/iso639-2/php/code_list.php

It is my understanding that the tags are correct - en maps to English and fr maps to French

Can you confirm that this still needs to be done?

kms6bn: @jeffreyliu3230 This is ready to merge, right?

jeffreyliu3230: Oh good find! Yes this should be ready to merge.

PR #523, opened 2016-04-15
Feature/add uwo

kms6bn:
JIRA SHARE-297


erinspace:
added in latest release

